PR #2, opened 2020-01-08
Add InternalDataCorruptionError

andrewjdawson2016:


thrift/shared.thrift
```diff
@@ -28,6 +28,10 @@ exception InternalServiceError {
   1: required string message
 }
 
+exception InternalDataCorruptionError {
```
venkat1109: should this be called DataCorruption or DataInconsistency ? Almost all of the cases we have are data inconsistency

andrewjdawson2016: Resolved

PR #4, opened 2020-01-15
Admin API to refresh all tasks of a workflow

anish531213:


thrift/history.thrift
```diff
@@ -857,4 +862,19 @@ service HistoryService {
...
+  /**
+  * RefreshWorkflowTasks refreshes all tasks of a workflow
+  **/
+  void RefreshWorkflowTasks(1: RefreshWorkflowTasksRequest request)
+    throws (
+      1: shared.BadRequestError badRequestError,
+      2: shared.InternalServiceError internalServiceError,
+      3: shared.EntityNotExistsError entityNotExistError,
+      4: ShardOwnershipLostError shardOwnershipLostError,
+      5: shared.LimitExceededError limitExceededError,
```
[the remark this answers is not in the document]
anish531213: Yeah...that's true. I will update that.

thrift/history.thrift
```diff
@@ -871,10 +871,8 @@ service HistoryService {
     throws (
       1: shared.BadRequestError badRequestError,
       2: shared.InternalServiceError internalServiceError,
-      3: shared.EntityNotExistsError entityNotExistError,
```
yycptt: I think this error still applies. If the API is called on a workflow that doesn't exist, this error should be returned. It's different from the no-op case when the workflow has already closed.

thrift/admin.thrift
```diff
@@ -141,6 +141,16 @@ service AdminService {
...
     )
+
+  /**
+  * RefreshWorkflowTasks refreshes all tasks of a workflow
+  **/
+  void RefreshWorkflowTasks(1: shared.RefreshWorkflowTasksRequest request)
+   throws (
+      1: shared.BadRequestError badRequestError,
+      2: shared.DomainNotActiveError domainNotActiveError,
+      3: shared.ServiceBusyError serviceBusyError,
```
yycptt: Please also add the `entityNotExistError` here.

anish531213: Done

thrift/admin.thrift
```diff
@@ -141,6 +141,17 @@ service AdminService {
...
       2: shared.ServiceBusyError serviceBusyError,
     )
+
+  /**
+  * RefreshWorkflowTasks refreshes all tasks of a workflow
+  **/
+  void RefreshWorkflowTasks(1: shared.RefreshWorkflowTasksRequest request)
+   throws (
+      1: shared.BadRequestError badRequestError,
+      2: shared.DomainNotActiveError domainNotActiveError,
```
yycptt: If this error is returned by the history client, also add it to `history.thrift`

PR #7, opened 2020-01-28
Add idl for admin API on DLQ

yux0:


thrift/admin.thrift
```diff
@@ -217,3 +247,40 @@ struct DescribeClusterResponse {
   10: optional shared.SupportedClientVersions supportedClientVersions
   20: optional MembershipInfo membershipInfo
 }
+
+enum QueueType {
```
yycptt: DLQType?

thrift/admin.thrift
```diff
@@ -217,3 +247,40 @@ struct DescribeClusterResponse {
...
+  Replication,
+  Domain,
+}
+
+struct ReadDLQMessagesRequest{
+  10: optional QueueType queueType
+  20: optional i32 shardID
```
yycptt: If the shardID doesn't exist, shall we return `EntityNotExistError`?

thrift/admin.thrift
```diff
@@ -217,3 +247,40 @@ struct DescribeClusterResponse {
...
+  30: optional i64 (js.type = "Long") inclusiveEndMessageID
+  40: optional i32 maximumPageSize
+  50: optional binary nextPageToken
+}
+
+struct ReadDLQMessagesResponse{
```
vancexu: Can we also add `QueueType` in response? so that in future we can extend this with 
```
11: optional list<visibilityTask> visibilityTasks
```

thrift/admin.thrift
yux0: Updated naming

thrift/admin.thrift
yux0: Correct. Added the EntityNotExistError

thrift/admin.thrift
```diff
@@ -217,3 +247,40 @@ struct DescribeClusterResponse {
...
+
+enum QueueType {
+  Replication,
+  Domain,
+}
+
+struct ReadDLQMessagesRequest{
+  10: optional QueueType queueType
+  20: optional i32 shardID
+  30: optional i64 (js.type = "Long") inclusiveEndMessageID
```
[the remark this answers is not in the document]
yux0: We have a track of the ack level on DLQ. So we don't need startMessageID and I don't see a case to read messages not from the beginning

thrift/admin.thrift
yux0: I think about this use datablob vs type + struct. I go with the later one b/c the history replication tasks are not raw events.

thrift/admin.thrift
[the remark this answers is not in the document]
vancexu: ok

thrift/admin.thrift
vancexu: ok

PR #9, opened 2020-02-04
Add DLQ APIs in history service for history replication DLQ

yux0:


thrift/history.thrift
```diff
@@ -857,4 +857,39 @@ service HistoryService {
...
+
+  /**
+  * MergeDLQMessages merges messages from DLQ
+  **/
+  replicator.MergeDLQMessagesResponse MergeDLQMessages(1: replicator.MergeDLQMessagesRequest request)
+    throws (
+      1: shared.BadRequestError badRequestError,
+      2: shared.InternalServiceError internalServiceError,
+      3: shared.ServiceBusyError serviceBusyError,
+      4: shared.EntityNotExistsError entityNotExistError,
```
yycptt: `ShardOwnerShipLostError`?

yux0: Good catch. Updated

thrift/history.thrift
```diff
@@ -857,4 +857,40 @@ service HistoryService {
       6: ShardOwnershipLostError shardOwnershipLostError,
       7: shared.EntityNotExistsError entityNotExistError,
     )
+
+  /**
+  * ReadDLQMessages returns messages from DLQ
+  **/
+  replicator.ReadDLQMessagesResponse ReadDLQMessages(1: replicator.ReadDLQMessagesRequest request)
```
venkat1109: Does it make sense to name these ReadReplicatorDLQMessages ? If there is any other type of DLQ in the future, then the naming becomes confusing.

yux0: My plan is to create a generic API for DLQ. The request contains a type to support multiple types.

venkat1109: Makes sense. In that case is the return type 'replicator.ReadDLQMessagesResponse' also generic ?

yux0: Yes. It also includes the DLQ type

PR #14, opened 2020-02-27
matching: add taskSource param to addTask APIs

venkat1109:
This patch adds a taskSource param to AddTask API request structs. The task source for now is either (1) history service or (2) matching db_backlog. This param will be used by the implementation of scalable task list i.e. when child partitions attempt remote-sync-match with parent partition, this info is needed to decide if certain operations can be performed.

### DIFF
--- a/thrift/matching.thrift
+++ b/thrift/matching.thrift
@@ -24,7 +24,7 @@ namespace java com.uber.cadence.matching
 
 // TaskSource is the source from which a task was produced
 enum TaskSource {
-    HISTORY    // Task produced by history service
+    HISTORY,    // Task produced by history service
     DB_BACKLOG // Task produced from matching db backlog
 }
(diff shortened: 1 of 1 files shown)


thrift/matching.thrift
```diff
@@ -22,6 +22,12 @@ include "shared.thrift"
 
 namespace java com.uber.cadence.matching
 
+// TaskSource is the source from which a task was produced
+enum TaskSource {
+    HISTORY    // Task produced by history service
```
vancexu: thrift enum may need a following comma at the end of value, see https://diwakergupta.github.io/thrift-missing-guide/

venkat1109: interesting. I think you are right - but it generated code just fine when I tested this without the comma.

PR #15, opened 2020-03-12
Add last error related information to ActivityTaskStartedEventAttributes

emrahs:
Adding the fields that are needed to expose information about retried activity task execution errors. See the related Cadence issue for more details: https://github.com/uber/cadence/issues/1873

We had 3 ways to make this change:
1) [Proposed] Creating two simple fields: lastFailureReason and lastFailureDetails.
2) Defining a new struct (ie. ActivityTaskError) that contains these error-related fields. In the future, we could add more details to this struct such as when the failure was hit, which event ID this failure corresponds to, etc., and reuse it in other places.
3) Having a list of structs in the ActivityTaskStartedEventAttributes event to make this even more extensible in case we want to expose "last N errors" instead of the "last" error. 

I like Option 3 more due to its extensibility but decided to go with Option 1 for sake of consistency with ActivityTaskTimedOutEventAttributes and PendingActivityInfo structs.


.gitignore
```diff
@@ -0,0 +1 @@
+.idea/
```
vitarb: What is that?

PR #26, opened 2020-04-14
Add rawHistorySupported flag for backward compatability

anish531213:
Added rawHistorySupported flag on GetWorkflowExecutionHistoryRequest to make backward compatibility with the client.


thrift/shared.thrift
```diff
@@ -1258,6 +1258,7 @@ struct GetWorkflowExecutionHistoryRequest {
   50: optional bool waitForNewEvent
   60: optional HistoryEventFilterType HistoryEventFilterType
   70: optional bool skipArchival
+  80: optional bool rawHistorySupported
```
mkolodezny, line 1261: nit: we might consider to name it just rawHistory and provide it as an option for client

mkolodezny:
Discussed offline: lets use the server version checker to check client instead.

PR #30, opened 2020-05-06
Add idl for admin API to do ndc workflow resend

yux0:


thrift/admin.thrift
```diff
@@ -179,12 +179,22 @@ service AdminService {
...
+    throws (
       1: shared.BadRequestError badRequestError,
       2: shared.DomainNotActiveError domainNotActiveError,
       3: shared.ServiceBusyError serviceBusyError,
       4: shared.EntityNotExistsError entityNotExistError,
-   )
+    )
+
+  /**
+  * ResendReplicationTasks sends replication tasks from remote cluster
```
yycptt, line 190: **to** remote cluster?

yux0: Updated the comment to request replication tasks from remote cluster and apply task to current cluster

PR #31, opened 2020-05-20
Adding failover marker replication task

yux0:


thrift/replicator.thrift
```diff
@@ -222,3 +224,10 @@ struct MergeDLQMessagesRequest{
 struct MergeDLQMessagesResponse{
   10: optional binary nextPageToken
 }
+
+struct FailoverMarkerAttributes{
```
yycptt, line 228: Probably move to L111 so that it's grouped with other replication task types

yux0: done

PR #34, opened 2020-06-01
Add RemoteSyncMatchedError type

vancexu:


thrift/shared.thrift
```diff
@@ -105,6 +105,10 @@ exception CurrentBranchChangedError {
   20: required binary currentBranchToken
 }
 
+exception RemoteSyncMatchedError {
+  10: required string message
```
mkolodezny, line 109: nit: do we need it to be required?

PR #42, opened 2020-09-10
Remove kafka related type

yux0:


yux0:
There could be backward compatibility issue.

PR #49, opened 2020-10-19
Add more fields to ActivityLocalDispatchInfo

mkolodezny:


yux0 (approved):

thrift/shared.thrift
```diff
@@ -406,7 +406,10 @@ struct ScheduleActivityTaskDecisionAttributes {
 
 struct ActivityLocalDispatchInfo{
   10: optional string activityId
-  // more fields will be added here later
+  20: optional i64 (js.type = "Long") scheduledTimestamp
+  30: optional i64 (js.type = "Long") startedTimestamp
+  40: optional i64 (js.type = "Long") scheduledTimestampOfThisAttempt
```
yycptt, line 411: I thought local dispatch only applies to the first attempt of an activity. If activity fails and is being retried, it will still go through matching? If that's the case, seems like we don't need the `scheduledTimestampOfThisAttempt` field? Or am I missing some context here?

mkolodezny: No, you are right. Decided to keep it to be fully compatible with the original activity dispatch. Will remove it later probably.

PR #53, opened 2021-03-03
Add definition for workflow shadower

yycptt:


thrift/shadower.thrift
```diff
@@ -0,0 +1,87 @@
...
+// THE SOFTWARE.
+
+namespace java com.uber.cadence.shadower
+
+include "shared.thrift"
+
+const string ShadowerLocalDomainName = "cadence-shadower"
+const string ShadowerTaskList = "cadence-shadower-tl"
+
+const string ShadowWorkflowName = "cadence-shadow-workflow"
```
yux0, line 28: Will this be used as a prefix of the workflow name?

yycptt: Good call. I forgot to add workflow ID suffix in the definition.

I am thinking `ShadowWorkflowName`, is the workflow name used by the worker to start the workflow. And probably `ShadowWorkflowIDSuffix = "-shadow-workflow"` is the suffix we add after domain name for the workflowID, so the workflowID will look like `<domainName>-shadow-workflow`. Thoughts?

yux0: Sounds good to me

yux0 (approved):

PR #58, opened 2021-03-25
Introduce delay start parameter

demirkayaender:
With this change StartWorkflowExecutionRequest will have the delay start
param so client can configure delay start

### DIFF
--- a/thrift/shared.thrift
+++ b/thrift/shared.thrift
@@ -1113,6 +1113,7 @@ struct StartWorkflowExecutionRequest {
   140: optional Memo memo
   141: optional SearchAttributes searchAttributes
   150: optional Header header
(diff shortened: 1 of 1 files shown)


thrift/shared.thrift
```diff
@@ -1113,6 +1113,7 @@ struct StartWorkflowExecutionRequest {
   140: optional Memo memo
   141: optional SearchAttributes searchAttributes
   150: optional Header header
+  160: optional i32 delayWorkflowStartSeconds
```
emrahs, line 1116: nit: I'd prefer not using the "workflow" word in the param name since it's kind of obvious. `delayStartSeconds` or 'delayedStartSeconds' sounds good enough to me.

emrahs (approved):

PR #62, opened 2021-04-13
Add persistence info to DescribeClusterResponse

mantas-sidlauskas:
This helps to switch on and off advanced queries in UI automatically

### DIFF
--- a/thrift/admin.thrift
+++ b/thrift/admin.thrift
@@ -252,15 +252,26 @@ struct MembershipInfo {
   30: optional list<RingInfo> rings
 }
 
+struct PersistenceSetting {
+  10: optional string key
+  20: optional string value
+}
+
+struct PersistenceFeature {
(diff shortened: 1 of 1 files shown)


just-at-uber (approved):

mkolodezny (approved):

thrift/admin.thrift
```diff
@@ -252,9 +252,15 @@ struct MembershipInfo {
   30: optional list<RingInfo> rings
 }
 
+struct PersistenceInfo {
+  10: optional string name
```
yycptt, line 256: Sorry I guess I left the same comment in the server side pr. Is it possible to rename this field so that in the future we can add persistence info for the main storage without causing confusion? Looks like for right now this name field is just for visibility persistence.

vytautas-karpavicius (approved):

PR #63, opened 2021-04-14
Expand workflowAlreadyCompletedError to other apis

demirkayaender:
**What changed?**
`WorkflowExecutionAlreadyCompletedError` was originally thought for SignalWorkflow, CancelWorkflow and TerminateWorkflow. To have a consistent behavior with consistent states from the other APIs, expanding this error to some other APIs.

See https://github.com/uber/cadence/pull/4123/ for more details.

Following functions support `EntityNotExistsError` but doesn't seem to need `WorkflowExecutionAlreadyCompletedError`. Please check the list below and let me know if we need to support the new error for some of those functions 
```
cadence.thrift:
- DescribeDomain
- ListDomains
- UpdateDomain
- DeprecateDomain
- GetWorkflowExecutionHistory ? 
- ListOpenWorkflowExecutions
- ListClosedWorkflowExecutions
- ListWorkflowExecutions
- ListArchivedWorkflowExecutions
- ScanWorkflowExecutions
- CountWorkflowExecutions
- QueryWorkflow
- DescribeWorkflowExecution
- DescribeTaskList
- ListTaskListPartitions


history.thrift:
- GetMutableState
- PollMutableState
- DescribeWorkflowExecution
- ReplicateEventsV2
- DescribeMutableState
- GetDLQReplicationMessages
- QueryWorkflow
- ReapplyEvents
- RefreshWorkflowTasks
- ReadDLQMessages
- PurgeDLQMessages
- MergeDLQMessages

matching.thrift
- RespondQueryTaskCompleted
- DescribeTaskList
```

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**

### DIFF
--- a/thrift/cadence.thrift
+++ b/thrift/cadence.thrift
@@ -155,6 +155,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
@@ -171,6 +172,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
@@ -207,6 +209,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
@@ -224,6 +227,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
@@ -241,6 +245,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
@@ -258,6 +263,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
@@ -275,6 +281,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
@@ -292,6 +299,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
@@ -309,6 +317,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
@@ -326,6 +335,7 @@ service WorkflowService {
       5: shared.LimitExceededError limitExceededError,
       6: shared.ServiceBusyError serviceBusyError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
     )
 
   /**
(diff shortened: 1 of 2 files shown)


thrift/history.thrift
```diff
@@ -644,6 +653,7 @@ service HistoryService {
       5: shared.DomainNotActiveError domainNotActiveError,
       6: shared.LimitExceededError limitExceededError,
       7: shared.ServiceBusyError serviceBusyError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
```
yycptt, line 656: I don't think ResetWorkflowExecution will return this error? It's ok to reset a completed workflow.

thrift/history.thrift
```diff
@@ -544,6 +550,7 @@ service HistoryService {
...
```
yycptt, line 553: Also need to update `RespondActivityTaskCompleted` above I think.

thrift/cadence.thrift
```diff
@@ -137,6 +138,7 @@ service WorkflowService {
       5: shared.EntityNotExistsError entityNotExistError,
       6: shared.DomainNotActiveError domainNotActiveError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
```
emrahs, line 141: Does this API really need this error?

thrift/cadence.thrift
```diff
@@ -190,6 +194,7 @@ service WorkflowService {
...
```
emrahs, line 197: Does this API really need this error?

thrift/cadence.thrift
```diff
@@ -391,6 +405,7 @@ service WorkflowService {
       5: shared.DomainNotActiveError domainNotActiveError,
       6: shared.LimitExceededError limitExceededError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
```
emrahs, line 408: Resetting a completed workflow is a valid use case. Is there a scenario where we might want to throw this error from Reset endpoint?

thrift/cadence.thrift
```diff
@@ -497,6 +512,7 @@ service WorkflowService {
       5: shared.ServiceBusyError serviceBusyError,
       6: shared.DomainNotActiveError domainNotActiveError,
       7: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
```
emrahs, line 515: Completed workflows can still be queried.

thrift/history.thrift
```diff
@@ -508,6 +513,7 @@ service HistoryService {
...
     )
 
   /**
```
emrahs, line 519: We should add the error to RespondActivityTaskCompleted, too

thrift/history.thrift
emrahs, line 656: Resetting a completed workflow is a valid use case.

thrift/history.thrift
```diff
@@ -508,6 +513,7 @@ service HistoryService {
       5: shared.DomainNotActiveError domainNotActiveError,
       6: shared.LimitExceededError limitExceededError,
       7: shared.ServiceBusyError serviceBusyError,
+      8: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
```
yycptt, line 516: This will cause some issues at client side I think. For record heartbeat, right now client side will check if the error type is entity not exist and cancel the activity if so.

thrift/history.thrift
```diff
@@ -456,6 +458,7 @@ service HistoryService {
       6: shared.DomainNotActiveError domainNotActiveError,
       7: shared.LimitExceededError limitExceededError,
       8: shared.ServiceBusyError serviceBusyError,
+      9: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
```
yycptt, line 461: This and `RecordDecisionTaskStarted` will also affect the error handling logic in matching.

thrift/cadence.thrift
```diff
@@ -108,6 +108,7 @@ service WorkflowService {
       6: shared.LimitExceededError limitExceededError,
       7: shared.EntityNotExistsError entityNotExistError,
       8: shared.ClientVersionNotSupportedError clientVersionNotSupportedError,
+      9: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
```
yycptt, line 111: Hmm, I think this should match the errors thrown by the history service. Same for SignalWithStartWorkflowExecution.

I understand that currently StartWorkflowExecution and SignalWithStartWorkflowExecution may return ErrAlreadyCompleted when workflowID reuse policy is TerminateIfRunning, but this looks like a bug to me, and those two APIs should not return entityNotExist or workflowAlreadyCompleted.

thrift/cadence.thrift
yycptt, line 141: In which case PollForDecision/ActivityTask will return this error?

thrift/cadence.thrift
yycptt, line 515: I am not sure about this one. @andrewjdawson2016 Can you double check?

yycptt (commented):
I am not sure if error handling for other APIs like RespondDecisionTaskCompleted need to updated at the Go client side. Java client side also not sure. 

But I guess my main concern here is that how do we release this change? This seems to be breaking change for the client and we are not able to control the client version our customer is using. Maybe we can increase the client side feature version and return the new error type only for the new feature version.

thrift/cadence.thrift
demirkayaender: Call stack for already completed error is as following: 
```
- Service/history/historyEngine.go StartWorkflowExecution
	- Service/history/historyEngine.go startWorkflowHelper
		- Service/history/historyEngine.go terminateAndStartWorkflow
			 - UpdateWorkflowLoop
			      if !runningMutableState.IsWorkflowExecutionRunning() {
			              return nil, workflow.ErrAlreadyCompleted
		              }
```

thrift/history.thrift
demirkayaender: Could you give more details? Could you point me to the places in matching if we need to address.

thrift/history.thrift
demirkayaender: I will send a client diff to handle this in `internal/internal_task_handlers.go:internalHeartBeat`

thrift/history.thrift
demirkayaender: never mind, found them, will update the other PR.

demirkayaender:
Per @yycptt's comment:

> I guess my main concern here is that how do we release this change? This seems to be breaking change for the client and we are not able to control the client version our customer is using. Maybe we can increase the client side feature version and return the new error type only for the new feature version.

This is a valid concern. How do we normally handle breaking changes on the client side?

thrift/cadence.thrift
emrahs: Seems like PollForDecisionTask should handle the error instead of passing it to the caller

thrift/cadence.thrift
emrahs: Same as PollForDecisionTask

thrift/cadence.thrift
emrahs: Seems like this would only happen in a race condition where a workflow is in running state when first checked but is no longer in that state by the time we are trying to terminate it. I think the right thing to do here is to handle the error and proceed as if it was completed when first checked.

emrahs:
We can't force all clients to upgrade before the server is updated, so the server has to be backward compatible. As @yycptt suggested, we should increment the feature version on the server, check the client version and return the old error if it's using the old client version.

demirkayaender:
> we should increment the feature version on the server, check the client version and return the old error if it's using the old client version.

is there any example to this?

emrahs:
Check this out: https://github.com/uber/cadence/pull/3199

emrahs (approved):

PR #71, opened 2021-05-27
Switch proto rpcs between services

vytautas-karpavicius:
A couple of RPCs were misplaced in the wrong service during initial api iteration.
- Move `RequestCancelWorkflowExecution` to `WorkflowService`. This is not part or the worker. User may cancel workflow execution from the client.
- Move `ResetStickyTaskList` to `WorkerService`. This is only used by the worker.

At this point we can still do this change. Clients are not yet released so we will not break them.

### DIFF
--- a/proto/uber/cadence/api/v1/service_worker.proto
+++ b/proto/uber/cadence/api/v1/service_worker.proto
@@ -136,11 +136,14 @@ service WorkerAPI {
   // API and return the query result to client as a response to 'QueryWorkflow' API call.
   rpc RespondQueryTaskCompleted(RespondQueryTaskCompletedRequest) returns (RespondQueryTaskCompletedResponse);
 
-  // RequestCancelWorkflowExecution is called by application worker when it wants to request cancellation of a workflow
-  // instance. It will result in a new 'WorkflowExecutionCancelRequested' event being written to the workflow history
-  // and a new DecisionTask created for the workflow instance so new decisions could be made. It fails with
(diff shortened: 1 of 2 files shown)


mkolodezny (approved):
LGTM

proto/uber/cadence/api/v1/service_worker.proto
```diff
@@ -136,11 +136,14 @@ service WorkerAPI {
   // API and return the query result to client as a response to 'QueryWorkflow' API call.
   rpc RespondQueryTaskCompleted(RespondQueryTaskCompletedRequest) returns (RespondQueryTaskCompletedResponse);
 
-  // RequestCancelWorkflowExecution is called by application worker when it wants to request cancellation of a workflow
-  // instance. It will result in a new 'WorkflowExecutionCancelRequested' event being written to the workflow history
-  // and a new DecisionTask created for the workflow instance so new decisions could be made. It fails with
-  // 'EntityNotExistsError' if the workflow is not valid anymore due to completion or doesn't exist.
-  rpc RequestCancelWorkflowExecution(RequestCancelWorkflowExecutionRequest) returns (RequestCancelWorkflowExecutionResponse);
```
mkolodezny, line 143: assuming its a backwards compatible

vytautas-karpavicius: In general it is not. However we have not released clients yet, so we can still change it.

PR #72, opened 2021-06-01
Change ListX Requests to Filter Cron Workflows

demirkayaender:
Changing List workflow requests so we can filter workflows based on whether they are cron or non-cron workflows

### DIFF
--- a/thrift/shared.thrift
+++ b/thrift/shared.thrift
@@ -1357,6 +1357,7 @@ struct ListOpenWorkflowExecutionsRequest {
   40: optional StartTimeFilter StartTimeFilter
   50: optional WorkflowExecutionFilter executionFilter
   60: optional WorkflowTypeFilter typeFilter
+  70: optional bool isCron
 }
 
 struct ListOpenWorkflowExecutionsResponse {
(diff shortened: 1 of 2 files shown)


thrift/shared.thrift
```diff
@@ -1384,6 +1386,7 @@ struct ListWorkflowExecutionsRequest {
   20: optional i32 pageSize
   30: optional binary nextPageToken
   40: optional string query
+  50: optional bool isCron
```
just-at-uber, line 1389: I'm not sure if this is needed here? Will this be part of query? Or is this the only way this will be able to be requested?

The reason asked is because it will impact the way it is implemented on the frontend. i.e. users will not be able to type in the textbox as part of the advanced query. It will likely need to be an accompanying dropdown.

yycptt: I think `isCron` should be part of the `query`, and we need to change how the query parser is implemented on the server side to support this.

demirkayaender: Server side changes are here: https://github.com/uber/cadence/pull/4242

Since this is optional, just introducing the field shouldn't change any behavior right? Later, frontend can set this field only when it's set by the user.  Is there something else we need to be careful about?

demirkayaender:
abandoning since we decided to go with ES only solution

PR #73, opened 2021-06-07
Add IDL changes to support feature flags

demirkayaender:
Introducing FeatureFlags and FeatureNotEnabledError in IDL repo. These changes will help us ship breaking changes without affecting customer workflows

Review/Land order:
- IDL changes: uber/cadence-idl#73
- Client-side changes: https://github.com/uber-go/cadence-client/pull/1103
- Server-side changes: uber/cadence#4257
- Samples changes: uber-common/cadence-samples#45

### DIFF
--- a/thrift/shared.thrift
+++ b/thrift/shared.thrift
@@ -96,6 +96,10 @@ exception ClientVersionNotSupportedError {
   3: required string supportedVersions
 }
 
+exception FeatureNotEnabledError {
+  1: required string featureFlag
+}
+
 exception CurrentBranchChangedError {
   10: required string message
   20: required binary currentBranchToken
@@ -1693,4 +1697,8 @@ struct ClusterInfo{
 struct RefreshWorkflowTasksRequest {
   10: optional string domain
   20: optional WorkflowExecution execution
+}
+
+struct FeatureFlags {
+	10: optional bool WorkflowExecutionAlreadyCompletedErrorEnabled
 }


thrift/shadower.thrift
```diff
@@ -71,6 +71,7 @@ struct ScanWorkflowActivityParams {
   30: optional binary nextPageToken
   40: optional i32 pageSize
   50: optional double samplingRate
+  60: optional shared.FeatureFlags featureFlags
```
yycptt, line 74: I don't think this is related? The struct is just the activity input for a system workflow.

thrift/shared.thrift
```diff
@@ -96,6 +96,10 @@ exception ClientVersionNotSupportedError {
   3: required string supportedVersions
 }
 
+exception FeatureNotEnabledError {
```
yycptt, line 99: Shall we use this error somewhere?

thrift/shadower.thrift
demirkayaender: See the client diff, I'm adding this flag to any place making rpc calls in case there will be a reason to turn on/off some features. 

More than that, I did it so we won't have to track which RPC calls would support this; by default it might be a good idea to enable this logic for all RPC calls. We can chat about this if you'd like.

thrift/shared.thrift
demirkayaender: This is used in server code currently. We might want to expose this to the user when they are requesting a specific behavior.

demirkayaender: yeah this is only until the client side changes go in

yycptt (approved):

PR #75, opened 2021-06-10
Add idl for fetching and responding cross cluster tasks

yycptt:
Proto definition change will be done in the server repo since they are only used by cadence server

### DIFF
--- a/thrift/shared.thrift
+++ b/thrift/shared.thrift
@@ -1723,6 +1723,7 @@ enum CrossClusterTaskType {
 
 enum CrossClusterTaskFailedCause {
   DOMAIN_NOT_ACTIVE
+  DOMAIN_NOT_EXISTS
   WORKFLOW_ALREADY_RUNNING
   WORKFLOW_NOT_EXISTS
 }
@@ -1734,8 +1735,7 @@ struct CrossClusterTaskInfo {
   40: optional CrossClusterTaskType taskType
   50: optional i16 taskState
   60: optional i64 (js.type = "Long") taskID
-  70: optional i64 (js.type = "Long") version
-  80: optional i64 (js.type = "Long") visibilityTimestamp
+  70: optional i64 (js.type = "Long") visibilityTimestamp
 }
 
 struct CrossClusterStartChildExecutionRequestAttributes {
--- a/thrift/shared.thrift
+++ b/thrift/shared.thrift
@@ -1789,7 +1789,7 @@ struct CrossClusterTaskRequest {
 struct CrossClusterTaskResponse {
   10: optional i64 (js.type = "Long") taskID
   20: optional CrossClusterTaskType taskType
-  30: optional CrossClusterTaskFailedCause failureCause
+  30: optional CrossClusterTaskFailedCause failedCause
   40: optional CrossClusterStartChildExecutionResponseAttributes startChildExecutionAttributes
   50: optional CrossClusterCancelExecutionResponseAttributes cancelExecutionAttributes
   60: optional CrossClusterSignalExecutionResponseAttributes signalExecutionAttributes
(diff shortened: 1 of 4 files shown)


thrift/shared.thrift
```diff
@@ -1713,4 +1713,100 @@ struct RefreshWorkflowTasksRequest {
...
+  WORKFLOW_ALREADY_RUNNING
+  WORKFLOW_NOT_EXISTS
+}
+
+struct CrossClusterTaskInfo {
+  10: optional string domainID
+  20: optional string workflowID
+  30: optional string runID
+  40: optional CrossClusterTaskType taskType
+  50: optional i16 taskState
```
yycptt, line 1735: Another options it to define new taskTypes is `CrossClusterTaskType` and add the corresponding task response/request attributes

thrift/shared.thrift
```diff
@@ -1713,4 +1713,100 @@ struct RefreshWorkflowTasksRequest {
...
+  10: optional string targetDomainID
+  20: optional string targetWorkflowID
+  30: optional string targetRunID
+  40: optional string requestID
+  50: optional i64 (js.type = "Long") initiatedEventID
+  60: optional bool childWorkflowOnly
+}
+
+struct CrossClusterCancelExecutionResponseAttributes { 
+}
```
yycptt, line 1762: Define empty struct here in case we want to return some information from the target cluster in the future. Same for `CrossClusterSignalExecutionResponseAttributes`

thrift/shared.thrift
```diff
@@ -1713,4 +1713,100 @@ struct RefreshWorkflowTasksRequest {
...
+  20: optional string workflowID
+  30: optional string runID
+  40: optional CrossClusterTaskType taskType
+  50: optional i16 taskState
+  60: optional i64 (js.type = "Long") taskID
+  70: optional i64 (js.type = "Long") version
+  80: optional i64 (js.type = "Long") visibilityTimestamp
+}
+
+struct CrossClusterStartChildExecutionRequestAttributes {
```
yycptt, line 1741: Another option for those RequestAttributes is to reuse the Request struct we defined for history Start/Signal/Cancel workflow API. But the problem with that is 1: it assumes how tasks are processed in the target cluster (just calling the corresponding API) 2: those history request structs contain some redundant fields which are not required for processing the task.

thrift/shared.thrift
```diff
@@ -1713,4 +1713,100 @@ struct RefreshWorkflowTasksRequest {
...
+}
+
+struct RespondCrossClusterTasksCompletedRequest {
+  10: optional i32 shardID
+  20: optional string targetCluster
+  30: optional list<CrossClusterTaskResponse> taskResponses
+}
+
+struct RespondCrossClusterTasksCompletedResponse {
+  10: optional list<CrossClusterTaskRequest> tasks
```
yycptt, line 1811: For returning new x-cluster task from the source cluster and save one x-cluster call.

thrift/shared.thrift
```diff
@@ -1713,4 +1713,100 @@ struct RefreshWorkflowTasksRequest {
...
+  50: optional i16 taskState
+  60: optional i64 (js.type = "Long") taskID
+  70: optional i64 (js.type = "Long") visibilityTimestamp
+}
+
+struct CrossClusterStartChildExecutionRequestAttributes {
+  10: optional string targetDomainID
+  20: optional string requestID
+  30: optional i64 (js.type = "Long") initiatedEventID
+  40: optional StartChildWorkflowExecutionInitiatedEventAttributes initiatedEventAttributes
```
yux0, line 1745: Do you need the run ID when the task is processed as create first decision in child workflow?

yux0 (approved):
LGTM. Have one question about start child workflow. Currently, we pass in the child run ID. I am not sure if we just create the decision in the current open workflow is fine.

thrift/shared.thrift
yycptt: Thanks for catching. I've added `targetRunID` to this struct and a comment saying it for scheduling the first decision.

yycptt:
Will land the diff after the server side change is ready.

PR #76, opened 2021-06-14
Add Cross Cluster Ops Related Structs

demirkayaender:


thrift/sqlblobs.thrift
```diff
@@ -241,6 +243,21 @@ struct TransferTaskInfo {
   32: optional i64 (js.type = "Long") visibilityTimestampNanos
 }
 
+struct CrossClusterTaskInfo {
```
yycptt, line 246: I think we can just reuse the `TransferTaskInfo` definition?

yycptt (approved):

PR #77, opened 2021-06-24
Add IDL for getTaskListByDomain API

yux0:


thrift/matching.thrift
```diff
@@ -109,6 +109,10 @@ struct DescribeTaskListRequest {
   20: optional shared.DescribeTaskListRequest descRequest
 }
 
+struct GetTaskListsByDomainRequest {
+  10: optional string domain
```
yycptt, line 113: Is this domainUUID or domainName? if is domainName, can we reuse the one defined in shared? If's UUID, shall we make it more clear in the naming?

yycptt (approved):

thrift/matching.thrift
yux0: It is domain name here

yux0: updated

thrift/cadence.thrift
```diff
@@ -576,6 +576,18 @@ service WorkflowService {
       2: shared.ServiceBusyError serviceBusyError,
     )
 
+  /**
+  * GetTaskListsByDomain returns the list of all the task lists for a domainName.
+  **/
+  shared.GetTaskListsByDomainResponse GetTaskListsByDomain(1: shared.GetTaskListsByDomainRequest request)
+    throws (
+      1: shared.BadRequestError badRequestError,
+      3: shared.EntityNotExistsError entityNotExistError,
```
yycptt, line 585: don't have to skip 2?

thrift/matching.thrift
yux0: updated

PR #78, opened 2021-07-02
Improve definition for cross-cluster related endpoints

yycptt:
- Add Workflow_Already_Completed and Uncategorized type for CrossClusterTaskFailed Cause
- Add per shard failed cause for getting cross cluster tasks
- Add fetchNewTasks bool field for indicating whether new tasks should be returned on responding task processing result

### DIFF
--- a/thrift/shared.thrift
+++ b/thrift/shared.thrift
@@ -1738,7 +1738,7 @@ enum CrossClusterTaskFailedCause {
   UNCATEGORIZED
 }
 
-enum GetCrossClusterTaskFailedCause {
+enum GetTaskFailedCause {
   SERVICE_BUSY
   TIMEOUT
   SHARD_OWNERSHIP_LOST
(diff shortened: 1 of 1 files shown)


thrift/shared.thrift
```diff
@@ -1734,6 +1734,15 @@ enum CrossClusterTaskFailedCause {
   DOMAIN_NOT_EXISTS
   WORKFLOW_ALREADY_RUNNING
   WORKFLOW_NOT_EXISTS
+  WORKFLOW_ALREADY_COMPLETED
+  UNCATEGORIZED
+}
+
+enum GetCrossClusterTaskFailedCause {
```
yycptt, line 1741: @yux0 Do you think replication stack may reuse this type? I can rename it to `GetTaskFailedCause` if so.

yux0: Yes. I think this is a good idea. Replication can use this in the future

yux0 (approved):

thrift/shared.thrift
yycptt: Sounds good. Renamed to `GetTaskFailedCause`

PR #82, opened 2021-08-19
Add Cross Cluster Parent Close Policy Types

demirkayaender:


thrift/shared.thrift
```diff
@@ -1808,12 +1808,24 @@ struct CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes {
 struct CrossClusterRecordChildWorkflowExecutionCompleteResponseAttributes {
 }
 
+struct CrossClusterApplyParentClosePolicyRequestAttributes {
+  10: optional string targetDomainID
+  20: optional string targetWorkflowID
+  30: optional string targetRunID
+  40: optional i64 (js.type = "Long") initiatedEventID
```
yux0, line 1815: What is this initiated event id used for?

yux0:
LGTM. We have two approaches to handle parent close policy, workflow or direct call. I assume we just support the direct call approach. Is this correct?

thrift/shared.thrift
```diff
@@ -1808,12 +1808,24 @@ struct CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes {
 struct CrossClusterRecordChildWorkflowExecutionCompleteResponseAttributes {
 }
 
+struct CrossClusterApplyParentClosePolicyRequestAttributes {
```
yycptt, line 1811: looks like this attribute is the information for applying parent close policy for one workflow. Does it mean we will generate N x-cluster tasks for N childworkflows in the source cluster, even if their target cluster is the same. Or there will be only one x-cluster task, but when sending the request to the target cluster, we represent them using N attributes?

thrift/shared.thrift
```diff
@@ -1808,12 +1808,27 @@ struct CrossClusterRecordChildWorkflowExecutionCompleteRequestAttributes {
 struct CrossClusterRecordChildWorkflowExecutionCompleteResponseAttributes {
 }
 
+struct AppyParentClosePolicyAttributes {
+  10: optional string domainID
+  20: optional string workflowID
+  30: optional string runID
```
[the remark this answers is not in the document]
demirkayaender: Yeah, target or child can make it easier to understand, let me change.

yycptt (approved):

PR #83, opened 2021-08-24
Add idl changes for get failover info API

yux0:


proto/uber/cadence/api/v1/domain.proto
```diff
@@ -63,6 +63,13 @@ message BadBinaryInfo {
   google.protobuf.Timestamp created_time = 3;
 }
 
+message FailoverInfo {
+  int64 failover_version = 1;
+  int64 failover_expired_timestamp = 2;
```
demirkayaender, line 68: is it possible to have failover_started_timestamp here?

yux0: We don't store this in DB. But I can calculate the start time based on the expired time. What do you think?

demirkayaender: I think it's a useful information especially in case we are going to show this in the UI as well. If it's going to be a too much work, we can talk about postponing this to a later time.

yux0: Yes. Added the start timestamp

thrift/history.thrift
```diff
@@ -351,6 +351,15 @@ struct DomainFilter {
   20: optional bool reverseMatch
 }
 
+struct GetFailoverInfoByDomainIDRequest {
+  10: optional string domainID
+}
+
+struct GetFailoverInfoByDomainIDResponse {
+  10: optional i32 completedShardCount
+  20: optional list<i32> pendingShards
```
demirkayaender, line 360: Not sure if we have this information either but, would be great to show how many requests remaining per shard for draining the traffic. 

This can be implemented later, not a blocker to this PR

demirkayaender (approved):

thrift/history.thrift
yux0: yeah. we don't have a good way to measure this. We can think about this later

demirkayaender: Do you mind creating a task and tracking this in the backlog? We can prioritize it later based on how much we need this information once you ship v0

proto/uber/cadence/api/v1/service_domain.proto
```diff
@@ -114,6 +114,7 @@ message DescribeDomainRequest {
 
 message DescribeDomainResponse {
   Domain domain = 1;
+  FailoverInfo failover_info = 2;
```
vytautas-karpavicius, line 117: Any reason why this is added to `DescribeDomainResponse` instead directly to `Domain`?
If we were to add it to `Domain`, other API call `ListDomains` would also get this additional info. Or is it intentionally left out there?

PR #84, opened 2021-08-25
Add taskState to CrossClusterTaskResponse

yycptt:
Since the execution for a cross cluster task has multiple stages(e.g. for signal execution, there's a signal step and a remove signal mutable state step) and the fact that source cluster may send duplicated x-cluster task requests, we need to know what's the stage the x-cluster task response is for.

### DIFF
--- a/thrift/shared.thrift
+++ b/thrift/shared.thrift
@@ -1819,11 +1819,12 @@ struct CrossClusterTaskRequest {
 struct CrossClusterTaskResponse {
   10: optional i64 (js.type = "Long") taskID
   20: optional CrossClusterTaskType taskType
-  30: optional CrossClusterTaskFailedCause failedCause
-  40: optional CrossClusterStartChildExecutionResponseAttributes startChildExecutionAttributes
-  50: optional CrossClusterCancelExecutionResponseAttributes cancelExecutionAttributes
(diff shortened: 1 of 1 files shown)


thrift/shared.thrift
```diff
@@ -1843,12 +1843,13 @@ struct CrossClusterTaskRequest {
 struct CrossClusterTaskResponse {
   10: optional i64 (js.type = "Long") taskID
   20: optional CrossClusterTaskType taskType
-  30: optional CrossClusterTaskFailedCause failedCause
-  40: optional CrossClusterStartChildExecutionResponseAttributes startChildExecutionAttributes
-  50: optional CrossClusterCancelExecutionResponseAttributes cancelExecutionAttributes
-  60: optional CrossClusterSignalExecutionResponseAttributes signalExecutionAttributes
-  70: optional CrossClusterRecordChildWorkflowExecutionCompleteResponseAttributes recordChildWorkflowExecutionCompleteAttributes
-  80: optional CrossClusterApplyParentClosePolicyResponseAttributes applyParentClosePolicyAttributes
+  30: optional i16 taskState
```
yycptt, line 1846: it's ok to make breaking changes now since the feature is still under development.

demirkayaender (approved):
looks good to me

PR #89, opened 2021-10-04
Add DomainList to TransferTaskInfo

demirkayaender:


thrift/sqlblobs.thrift
```diff
@@ -241,6 +241,7 @@ struct TransferTaskInfo {
   28: optional i64 (js.type = "Long") scheduleID
   30: optional i64 (js.type = "Long") version
   32: optional i64 (js.type = "Long") visibilityTimestampNanos
+  34: optional list<string> domainList
```
yycptt, line 244: nit: can we rename this field to something like `targetDomainIDList` or `targetDomainIDs`?

yycptt (approved):

PR #94, opened 2021-10-29
Compile go proto types

vytautas-karpavicius:
Generate public API proto types in this repository so that they could be shared among client and server.

Current setup (client and server generating their own copies) does not work. When server includes go sdk, both tries to register same gogo types which results in duplicate registrations. They happen in global `init()` blocks.

To prevent this, it is recommended to use common go package in both of them. This repository is a good candidate to store such common generated types.

Another benefit - we automatically get regenerated types once IDLs change.


go.sum
```diff
@@ -0,0 +1,295 @@
...
+github.com/beorn7/perks v0.0.0-20180321164747-3a771d992973/go.mod h1:Dwedo/Wpr24TaqPxmxbtue+5NUziq4I4S80YR8gNf3Q=
+github.com/beorn7/perks v1.0.0/go.mod h1:KWe93zE9D1o94FZ5RNwFwVgaQK1VOXiVxmqh+CedLV8=
+github.com/beorn7/perks v1.0.1 h1:VlbKKnNfV8bJzeqoa4cOKqO6bYr3WgKZxO8Z16+hsOM=
+github.com/beorn7/perks v1.0.1/go.mod h1:G2ZrVWU2WbWT9wwq4/hrbKbnv/1ERSJQ0ibhJ6rlkpw=
+github.com/bmizerany/perks v0.0.0-20141205001514-d9a9656a3a4b h1:AP/Y7sqYicnjGDfD5VcY4CIfh1hRXBUavxrvELjTiOE=
+github.com/bmizerany/perks v0.0.0-20141205001514-d9a9656a3a4b/go.mod h1:ac9efd0D1fsDb3EJvhqgXRbFx7bs2wqZ10HQPeU8U/Q=
+github.com/cactus/go-statsd-client/statsd v0.0.0-20191106001114-12b4e2b38748/go.mod h1:l/bIBLeOl9eX+wxJAzxS4TveKRtAqlyDpHjhkfO0MEI=
+github.com/census-instrumentation/opencensus-proto v0.2.1/go.mod h1:f6KPmirojxKA12rnyqOA5BBL4O983OfeGPqjHWSTneU=
+github.com/cespare/xxhash/v2 v2.1.1 h1:6MnRN8NT7+YBpUIWxHtefFZOKTAPgGjpQSxqLNn0+qY=
+github.com/cespare/xxhash/v2 v2.1.1/go.mod h1:VGX0DQ3Q6kWi7AoAeZDth3/j3BFtOZR5XLFGgcrjCOs=
```
Groxx, line 21: unfortunately this breaks glide and dep :\
looks like it's coming from yarpc -> prometheus, but it'll take a bit of fiddling to find out exactly where.

Groxx:
pull in the M1 fixes from https://github.com/uber/cadence/pull/4576 too please :)

go.sum
Groxx: Actually, this may be fine.  It's in our go.mod currently on 0.18.3, but dep doesn't pull it in at all, and it succeeds.

So meh, likely not an issue.  Dependencies look fine to me.

yux0:
I like the idea. One thing I am unclear about is breaking changes. Even though I don't think it will be an issue, will the release process stay the same as now?

Groxx:
> I like the idea. One thing I am unclear about is breaking changes. Even though I don't think it will be an issue, will the release process stay the same as now?

Since this is just grpc and I don't believe we support grpc in the client's APIs yet (they're all Thrift-based): seems fine to me.
But if we were to switch thrift to this, yeah - it'd be a different import path, which would be significantly breaking.

vytautas-karpavicius:
> > I like the idea. One thing I am unclear about is breaking changes. Even though I don't think it will be an issue, will the release process stay the same as now?
> 
> Since this is just grpc and I don't believe we support grpc in the client's APIs yet (they're all Thrift-based): seems fine to me. But if we were to switch thrift to this, yeah - it'd be a different import path, which would be significantly breaking.

It is breaking for people that started using gRPC via compatibility layer. This is rather unfortunate even though fix is straightforward - updating one import path.

I would argue that we should still do it. Rather earlier that later.

Another alternative - do not move generated types here, but import generated public API types in the server instead of generating them there. This would avoid this breakage as client stays the same.
However, this would mean that every time we update IDLs, we need to update client in the server basically to latest master as well. This seems a higher maintenance option and potentially bring other issues with unreleased client.

If you have other ideas - let me know.

vytautas-karpavicius:
> pull in the M1 fixes from [uber/cadence#4576](https://github.com/uber/cadence/pull/4576) too please :)
> 
> otherwise LGTM, though we lack the copyright header (does that even matter?), and I haven't checked for any other differences in the generated folder.

Added the M1 fix.

Regarding license headers - do we really need those here? Those are generated files, and we have headers on input IDLs.

Groxx:
tbh I have no idea.  copyright headers on files feels to me like legal cargo-culting that probably doesn't actually change anything.

Groxx (approved):
👍 sgtm.
I may have to go over the makefile carefully some time here, but it's fine for now.  it's much simpler than the client or server makefile, so it's also a lot easier to spot issues and fix them while developing.

PR #99, opened 2022-01-26
Admin Delete Workflow RPC and request

demirkayaender:


thrift/admin.thrift
```diff
@@ -251,6 +251,11 @@ service AdminService {
     throws (
       1: shared.InternalServiceError internalServiceError,
     )
+
+  void DeleteWorkflow(1: AdminDeleteWorkflowRequest request)
+    throws (
+      1: shared.InternalServiceError internalServiceError,
```
emrahs, line 257: I haven't seen the code that depends on this, yet, but I think we should also have the `badRequestError` and `entityNotExistError` here as user errors. I'd like to reduce the number of scenarios where `internalServiceErrors` are returned so that they are always a clear signal for bugs in our system that need to be fixed.

thrift/admin.thrift
```diff
@@ -359,6 +364,12 @@ struct RestoreDynamicConfigRequest {
   20: optional list<config.DynamicConfigFilter> filters
 }
 
+struct AdminDeleteWorkflowRequest {
+  10: optional string                       domain
+  20: optional shared.WorkflowExecution     execution
+  30: optional bool                         skipErrors
```
emrahs, line 370: What's `skipErrors` for? Perhaps rename it to `force` if that's inline with the intent.

vytautas-karpavicius (commented):
Please add corresponding protobuf changes as well

vytautas-karpavicius:
> Please add corresponding protobuf changes as well

Ah, nevermind... It is not in this repository.

thrift/admin.thrift
demirkayaender: I have been using the same parameter names from admin local delete command. Since that name is also exposed to API, it's probably better to use the same name.

thrift/admin.thrift
```diff
@@ -251,6 +251,20 @@ service AdminService {
     throws (
       1: shared.InternalServiceError internalServiceError,
     )
+
+  void DeleteWorkflow(1: AdminDeleteWorkflowRequest request)
+    throws (
+      1: shared.BadRequestError         badRequestError,
+      2: shared.EntityNotExistsError    entityNotExistError,
+      3: shared.InternalServiceError    internalServiceError,
```
davidporter-id-au, line 259: nit: please remove 5XX type errors from expected error defintions as there's no case where this is a valid state. In practice I realised this an admin API and it shouldn't matter, but it makes tracking down errors much more difficult.

demirkayaender: This logic is shared across our thrift files. If I remember this correctly this something to do with our Java client where it needs to know which type of errors are being thrown and InternalServiceError is the wrapper for all the undefined errors.

emrahs (approved):

PR #102, opened 2022-02-22
add refresh tasks API to workflow service

mkolodezny:
Expose Admin Task Refresh API in client SDK


Groxx (approved):
tbh I generally lean towards pushing people to make their own admin clients for this kind of thing....
... but it's not like it's harmful to expose.  and thrift's kinda on its way out, so there's no long upkeep needs really.

PR #112, opened 2022-06-27
Adding QOS controls workflow start

davidporter-id-au:
The idea is that service owners will probably best placed to indicate whether their workflow is latency sensitive or not, and workflow level feels likely the kind of granularity at which these controls should be implemented. Domains often have a mixture of both latency sensitive and batch content.

At the moment these changes do nothing, but the intent is to use them to inform decisions about load-shedding and rate-limiting going forward.


davidporter-id-au:
Not going with this after some discussion.

PR #115, opened 2022-07-05
Cdnc 1781

allenchen2244:
Add idl stuff for restartworkflow api


proto/uber/cadence/api/v1/service_workflow.proto
```diff
@@ -90,6 +95,10 @@ service WorkflowAPI {
   rpc RefreshWorkflowTasks(RefreshWorkflowTasksRequest) returns(RefreshWorkflowTasksResponse);
 }
 
+message RestartWorkflowExecutionRequest {
```
vytautas-karpavicius, line 98: I assume we are adding a new IDL because we want atomicity of such operation. That is, it is not enough to combine Start/Terminate as separate operations?

If that is the case, I imagine that restart should start the workflow exactly the same as it was started. Therefore start request here is not really needed. You only need fields that are present in terminate request. This alternatively could be a simply flag in `restart = true` in terminate request. All start parameters can (and should) be taken from history of existing workflow.

Otherwise, with current contract we can end up that workflow with on ID is terminated and started with another ID. This not longer seems like a restart to me.

allenchen2244: I think the goal is actually to just outright terminate and restart on a new workflow ID but let me confirm that.

PR #118, opened 2022-07-12
Cdnc 1781 v3. Adding an API that allows a simple restart or terminate then restart if there is an active runID

allenchen2244:
IDL for restart workflow api


proto/uber/cadence/api/v1/service_workflow.proto
```diff
@@ -90,6 +93,11 @@ service WorkflowAPI {
   rpc RefreshWorkflowTasks(RefreshWorkflowTasksRequest) returns(RefreshWorkflowTasksResponse);
 }
 
+message RestartWorkflowExecutionRequest {
+  string domain = 1;
+  WorkflowExecution workflow_execution = 2;
+  string identity = 3;
```
davidporter-id-au, line 99: What is identity?

thrift/cadence.thrift
```diff
@@ -94,6 +94,20 @@ service WorkflowService {
     )
 
   /**
+  * RestartWorkflowExecution restarts a previous workflow
+  * If the workflow is currently running it will terminate and restart
+  **/
+  shared.StartWorkflowExecutionResponse RestartWorkflowExecution(1: shared.RestartWorkflowExecutionRequest restartRequest)
+    throws (
+      1: shared.BadRequestError badRequestError,
+      3: shared.WorkflowExecutionAlreadyStartedError sessionAlreadyExistError,
```
davidporter-id-au, line 103: I think `2:` might be missing here. But overall, I think these are all 4XX errors, nice!

davidporter-id-au:
Please include more information in the title and description about what you're trying to achieve for future folks, particularly given the risk of ambiguity with this API

proto/uber/cadence/api/v1/service_workflow.proto
allenchen2244: identity is the guy making the call. it gets passed in in all requests pretty much

thrift/cadence.thrift
allenchen2244: woops will fix

proto/uber/cadence/api/v1/service_workflow.proto
davidporter-id-au: ah, sorry, I was expecting AuthN information to be passed as a header. But, as you say, this seems to be a field throughout the rest of the APIs. ok, fair.

davidporter-id-au (approved):
lgtm, maybe if we could get a second pair of eyes that might be good, just because IDL is one of those things that is tricky to change later on.

thrift/cadence.thrift
```diff
@@ -94,6 +94,19 @@ service WorkflowService {
     )
 
   /**
+  * RestartWorkflowExecution restarts a previous workflow
+  * If the workflow is currently running it will terminate and restart
+  **/
+  shared.StartWorkflowExecutionResponse RestartWorkflowExecution(1: shared.RestartWorkflowExecutionRequest restartRequest)
```
Groxx, line 100: using `Start*`, not `Restart*` response

proto/uber/cadence/api/v1/service_workflow.proto
```diff
@@ -90,6 +93,11 @@ service WorkflowAPI {
   rpc RefreshWorkflowTasks(RefreshWorkflowTasksRequest) returns(RefreshWorkflowTasksResponse);
 }
 
+message RestartWorkflowExecutionRequest {
```
Groxx, line 96: if the semantics we're going for are always "current run is terminated", we probably need to support either request-ID or an "only if current" / "only if current == UUID" argument.

otherwise, with a successful-but-timed-out request, you risk restarting a *second* time, and not having a good way to prevent that.

allenchen2244: what does this mean? "with a successful-but-timed-out request, you risk restarting a second time, and not having a good way to prevent that."

Right now it calls startworkflow with the policy terminate if the wfid has something running.

PR #119, opened 2022-07-20
Add update time to workflow execution info

neil-xie:
Added update time field to workflow execution info.

Explanation: Currently the visibility record only has start/close time fields, the updateTime field is added to keep track of the last time of update workflow execution, so we can know if they are not making progress for a long while. There are multiple steps to complete this goal, adding the UpdateTime field is the first step.


davidporter-id-au:
@neil-xie  can you add an explanation as to what this diff is trying to accomplish?

neil-xie:
> 



> @neil-xie can you add an explanation as to what this diff is trying to accomplish?

Sure! Let me update the description now

davidporter-id-au (approved):

davidporter-id-au:
I don't really have enough context on this diff. It looks good to me, but please get stamp with someone who's more familiar with this project maybe?

mkolodezny (commented):
LGTM

mkolodezny (approved):

PR #125, opened 2022-10-07
Adding domain to WorkflowExecution for cross domain signal calls.

mindaugasbarcauskas:
this is how it will be used in java-client:

WorkflowExecution workflowExecution =
    new WorkflowExecution().setWorkflowId(workflowId).setDomain("testdomain2");
ExternalWorkflowStub workflowStub =
    Workflow.newUntypedExternalWorkflowStub(workflowExecution);
workflowStub.signal("FareWellWorkflow::waitForName", "Mindaugas");
return Arrays.asList("FareWell workflow scheduled");


thrift/shared.thrift
```diff
@@ -368,6 +368,7 @@ struct TaskListMetadata {
 struct WorkflowExecution {
   10: optional string workflowId
   20: optional string runId
+  30: optional string domain
```
davidporter-id-au, line 371: I might need a little more context on this, in the go library this works by scheduling a decision task response with the domain filled in: https://github.com/uber/cadence-idl/blob/master/proto/uber/cadence/api/v1/decision.proto#L143

But possibly I'm not understanding what you're trying to do

PR #128, opened 2023-01-30
Add missing reason field to proto's RestartWorkflow API

Groxx:
Seems like it just got missed in #118.
It exists in the thrift version, just not in the proto:
https://github.com/uber/cadence-idl/blob/6c250ae4198419b2fb01aded7a4f9830ae705baa/thrift/shared.thrift#L1360-L1365


davidporter-id-au (approved):
lgtm but I'm not super privvy to the intent behind all the parts to this API

PR #130, opened 2023-02-07
Add PartitionConfig to requests definition

Shaddoll:
# Context
We want to support workflow zonal isolation in a regional Cadence cluster. 
We're introducing isolation groups to task list pollers to let pollers of the same task lists but from different isolation groups to poll tasks only from their isolation group.
We're introducing partition config to workflow and activity to let them be partitioned to a certain isolation group which will only be picked up by pollers from that isolation group.


thrift/shared.thrift
```diff
@@ -399,6 +399,7 @@ struct WorkflowExecutionInfo {
   120: optional string taskList
   130: optional bool isCron
   140: optional i64 (js.type = "Long") updateTime
+  150: optional map<string, string> partitionConfig
```
Shaddoll, line 402: This type is used by `DescribeWorkflowExecution`. Adding the config to expose it to customers.

davidporter-id-au (approved):

PR #133, opened 2023-03-11
Isolation groups API CRUD

davidporter-id-au:
Adds an interface for updating / reading isolation groups, both at the domain and globally. 

Eg (pseudocode)

```
  client.UpdateGlobalIsolationGroups(ctx, adminv1.UpdateDomainIsolationGroupsRequest{
      domain: "sample-domain",
     // a list of all the zones this is to not be operating in:
      isolation_groups: {
        { 
           name: "zone1",
           state:  adminv1.IsolationGroupStateDrained
        },
      }
  })
```

I wanted to be able to hold-off on an API which is clearly drain oriented or allow-oriented. The IsolationGroup implementation can determine to be deny or allow oriented in its use of this information. 

### Upsert semantics

By a simple GET and SET api I expect callers to add the total state they wish to see every time, this is a complete upsert. 

I expect a subsequent call such as this: 
```
  client.UpdateGlobalIsolationGroups(ctx, adminv1.UpdateDomainIsolationGroupsRequest{
      domain: "sample-domain",
      isolation_groups: {}
  })
```
to remove all drains for this domain


thrift/admin.thrift
```diff
@@ -265,6 +265,26 @@ service AdminService {
       2: shared.EntityNotExistsError    entityNotExistError,
       3: shared.InternalServiceError    internalServiceError,
     )
+
+  GetGlobalIsolationGroupsRequest GetGlobalIsolationGroups(1: GetGlobalIsolationGroupsResponse request)
```
Shaddoll, line 269: response <-> request

proto/uber/cadence/admin/v1/service.proto
```diff
@@ -435,3 +447,33 @@ message DynamicConfigFilter {
 	string name = 1;
 	api.v1.DataBlob value = 2;
 }
+
+// Global
+message GetGlobalIsolationGroupsRequest{}
+
+message GetGlobalIsolationGroupsResponse{
+    repeated IsolationGroupConfiguration isolation_groups = 1;
```
Shaddoll, line 455: should `repeated` be removed?

proto/uber/cadence/admin/v1/service.proto
```diff
@@ -121,6 +121,18 @@ service AdminAPI {
 
   // MaintainCorruptWorkflow deletes a workflow if its history is corrupt due to underlying DB issues (e.g. Cassandra resurrections)
   rpc MaintainCorruptWorkflow(MaintainCorruptWorkflowRequest) returns (MaintainCorruptWorkflowResponse);
+
+  // GetIsolationGroups
+  rpc GetGlobalIsolationGroups(GetGlobalIsolationGroupsRequest) returns (GetGlobalIsolationGroupsResponse);
```
Shaddoll, line 126: Who will call the `Get` APIs?

proto/uber/cadence/admin/v1/service.proto
davidporter-id-au: yes, thank you, I didn't notice I'd done that twice

thrift/admin.thrift
davidporter-id-au: thank you

proto/uber/cadence/admin/v1/service.proto
davidporter-id-au: CLI / SRE tooling. I doubt it's possible to have such an endpoint to add drains without being able to read also

Shaddoll (approved):

PR #136, opened 2023-03-26
Add thriftrw codegen, modernize makefile a bit

Groxx:
In a nutshell, this:
- moves codegen-only tools into internal/tools/go.mod, like client and server
- makes the makefile optionally verbose, like client and server
- adds thrift codegen, so we can get rid of the submodule in other repos
- lightly cleans up codegen and build rules to simplify and handle the additional files


davidporter-id-au:
> In a nutshell, this:
> 
> * moves codegen-only tools into internal/tools/go.mod, like client and server
> * makes the makefile optionally verbose, like client and server
> * adds thrift codegen, so we can get rid of the submodule in other repos
> * lightly cleans up codegen and build rules to simplify and handle the additional files

Not entirely sure I understand the plan/intention with the commmited thrift files? This doesn't look like all of them.

Groxx:
> Not entirely sure I understand the plan/intention with the commmited thrift files? This doesn't look like all of them.

it's all of the files that the client uses: https://github.com/uber-go/cadence-client/blob/master/Makefile#L188https://github.com/uber-go/cadence-client/blob/5ead82442dd272b8077d8eeb7d1b4c490c621c8c/Makefile#L188
if we need more in the server, yea - adding them makes sense.

I can update the commit message, but the intent is to get rid of .gen in the client (breaking change, will be part of v2), and to make the client and server share the same codegen, like we do for proto.

PR #142, opened 2023-05-02
Bugfix/zonal isolation correct persistence idl ii

davidporter-id-au:
- Pulls isolation-group entity structs into common protobuf folder
- Exposes isolation-group entity-strucs  in domain configuration struct in thrift
- exposes isolation-group entity structs in Domain struct in protobuf (it doesn't seem to have a configuration struct)


proto/uber/cadence/api/v1/domain.proto
```diff
@@ -48,6 +49,7 @@ message Domain {
   int64 failover_version = 15;
   bool is_global_domain = 16;
   FailoverInfo failover_info = 17;
+  api.v1.IsolationGroupConfiguration isolation_groups = 18;
```
Shaddoll, line 52: this type is in the same package, can we remove `api.v1.`?

Shaddoll (approved):

PR #143, opened 2023-05-04
Feature/zonal isolation

davidporter-id-au:
This adds endpoints and structures for the zonal-isolation feature: The ability to subdivide an operating region into more resilient subsections where workflows are confined when they are started. 

Internal to Uber and in some companies an 'isolation-group' is sometimes referred to a ['zone'](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/using-regions-availability-zones.html), but we are drawing a distinction here, since Cadence can operate sub-zonally with these too.

Workflows are expected to be tagged with the isolation-group they started with and can be partitioned according to whatever partitioning scheme is desired, in the default implementation this will be accomplished by pinning the workflows to the start isolation-group where they were created. 

Isolation-group-status:
A enum flag indicating whether an isolation-group is drained or not. For the sake of the initial open-source implementation, only the drained status is important, as anything not specified is expected to be healthy, however, the IDL offers the base for alternative implementations.

Isolation-group-partition:
A single isolation group, with it's name, it's status and potentially any future group-specific information.

isolation-group configuration
An entire set of isolation-groups. This is presented externally as an unsorted list of isolation-group partitions. 

### APIs for performing actions on Isolation-groups

Apart from engaging in the aforementioned 'pinning' behaviour mentioned above, this PR introduces the ability to perform 'drain' operations on these isolation-groups 

This PR also adds the following CRUD endpoints for operating on zones on both a domain and 'globally' (in the sense that it affects all domains). 

The APIs are Get / Set apis with upsert semantics. Ie (in pseudocode): 

```golang
// drain zone-1
client.UpdateGlobalIsolationGroups(ctx, adminv1.UpdateGlobalIsolationGroupsRequest{
  IsolationGroups: []*apiv1.IsolationGroupsConfiguration{
      {Name: "zone-1",  Status: apiv1.ISOLATION_GROUP_STATE_DRAINED},
   }
})

// remove all drains
client.UpdateGlobalIsolationGroups(ctx, adminv1.UpdateGlobalIsolationGroupsRequest{
  IsolationGroups: []*apiv1.IsolationGroupsConfiguration{}
})
```


thrift/admin.thrift
```diff
@@ -265,6 +265,26 @@ service AdminService {
       2: shared.EntityNotExistsError    entityNotExistError,
       3: shared.InternalServiceError    internalServiceError,
     )
+
+  GetGlobalIsolationGroupsResponse GetGlobalIsolationGroups(1: GetGlobalIsolationGroupsRequest request)
+    throws (
+      1: shared.BadRequestError badRequestError,
```
Shaddoll, line 271: I think we should also include `shared.InternalServiceError`, and I'm not sure if there is any other error which should be included.

davidporter-id-au: No, that's always an error. There's no valid scenario where returning an internal-service-error is expected. It is, by definition, an unexpected error. 

Adding it to the list here is to imply that this is expected when it isn't. This, regrettably was an error made through a lot of Uber services earlier and there's been a lot of work to undo.

Shaddoll (approved):

PR #147, opened 2023-07-26
[history] Sync internal WorkflowExecutionInfo with thrift shared

3vilhamster:
Adding fields for WorkflowExecutionInfo so we can have roundtrip conversion between internal types WorkflowExecutionInfo and thrift WorkflowExecutionInfo.
These fields exists in proto, but is missing in thrift.


gazi-yestemirova (approved):

davidporter-id-au:
This will probably need some codegen too

3vilhamster:
what do you mean by codegen?

PR #148, opened 2023-08-28
Add new fields of history size into pollForDecisionTaskResponse

timl3136:
In order to transmit history size gathered in cadence server to cadence client, we need to update fields in pollForDecisionTaskResponse to sync up thrift and proto fields between those two repos.

The history size is exposed in terms of byte and is the total history size of a specific workflow

The code is tested through unit test as well as testing in development environment.


proto/uber/cadence/api/v1/service_worker.proto
```diff
@@ -169,6 +169,7 @@ message PollForDecisionTaskResponse {
   google.protobuf.Timestamp started_time = 13;
   map<string, WorkflowQuery> queries = 14;
   int64 next_event_id = 15;
+  int64 history_size = 16;
```
Groxx, line 172: from chat: since this is the cumulative number of bytes in the workflow, `total_history_bytes` or something seems a lot more descriptive

timl3136: renaming finished

Groxx (approved):

PR #149, opened 2023-09-01
Sync fields between cadence-server and cadence-idl

timl3136:
Add fields in thrift for both GetMutableStateResponse and RecordDecisionTaskStartedResponse in sync with cadence-server.
Resolves the field mismatch in mapper due to discrepancy between thrift struct and go.


thrift/history.thrift
```diff
@@ -216,6 +217,7 @@ struct RecordDecisionTaskStartedResponse {
   120: optional i64 (js.type = "Long") scheduledTimestamp
   130: optional i64 (js.type = "Long") startedTimestamp
   140: optional map<string, shared.WorkflowQuery> queries
+  150: optional i64 (js.type = "Long") historySize
```
Shaddoll, line 220: should the name be totalhistorybytes?

timl3136: This is the field sent from cadence history to cadence matching, so I was thinking of keeping the same name used in persistence where we calculate the history size.

neil-xie (approved):

PR #153, opened 2023-11-06
Add user management API

mantas-sidlauskas:
Adding Proto and Thrift IDL for Cadence-native user management.


thrift/user.thrift
```diff
@@ -0,0 +1,66 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+namespace java com.uber.cadence.user
+
+struct User {
+    10: optional string username
+    20: optional bool admin
```
davidporter-id-au, line 25: Roles have a way of becoming more complex than bool. I'd suggest making this an enum, if you want to do RBAC type authentication

mantas-sidlauskas: I'm following current model in Cadence: admin and list of free-form groups. But making admin as a defined role is a good idea. Will update this PR

mantas-sidlauskas:
Closing this, we are planning to provide 3rd party Identity Provider for OAuth, this is not needed

PR #155, opened 2024-01-16
Define async workflow queue configuration for domains

taylanisikdemir:
**What changed?**
- `AsyncWorkflowConfiguration` object is defined in both thrift and proto files
- Get and Update methods exposed on AdminAPI

**Why?**
We need somewhere to persist async workflow queue configuration of a domain in order to support the upcoming async APIs. These APIs will absorb the burst of workflow initiation calls by buffering requests on a queue (e.g. kafka) and consume from there asynchronously.


proto/uber/cadence/api/v1/common.proto
```diff
@@ -124,4 +124,27 @@ message IsolationGroupPartition {
...
\ No newline at end of file
+}
+
+message AsyncWorkflowConfiguration {
+  // predefined_queue_name is the name of the predefined queue in cadence server config's asyncWorkflowQueues
+  string predefined_queue_name = 1;
+
+  // Below fields are only used if predefined_queue_name is not set
+  AsyncWorkflowQueueType queue_type = 2;
+  AsyncWorkflowKafkaQueueConfiguration kafka_config = 3;
```
Shaddoll, line 135: Should it be moved into a oneof block?

taylanisikdemir: Make sense, updated. I missed because I started with writing thrift struct and copied over to proto. Thrift doesn't have such concept and looks like we just declare every field as optional. See HistoryEvent for example.

Shaddoll (approved):

PR #157, opened 2024-01-22
Introduce StartWorkflowExecutionAsync API

Shaddoll:
**What changed?**

- Introduce StartWorkflowExecutionAsync API
- Introduce a new type to represent message in async queue

**Why?**

**How did you test it?**

**Potential risks**

**Release notes**

**Documentation Changes**


thrift/sqlblobs.thrift
```diff
@@ -276,3 +276,15 @@ struct ReplicationTaskInfo {
...
+  StartWorkflowExecutionAsyncRequest
+}
+
+struct AsyncRequestMessage {
+  10: optional string partitionKey
+  12: optional AsyncRequestType type
+  14: optional shared.Header header
+  16: optional string encoding
+  18: optional binary payload
+}
```
taylanisikdemir, line 290: what is the value of defining storage layer blobs in this idls repo? clients don't need to know about this.

taylanisikdemir (approved):

thrift/sqlblobs.thrift
Shaddoll: the kafka producer needs to encode the payload, and currently it uses thrift encoder in the codebase
https://github.com/uber/cadence/blob/master/common/messaging/kafka/producer_impl.go#L83

taylanisikdemir: Discussed offline. Basically cadence-idl repo is a place for shared proto/thrift files which both server and client sdk needs. Storage layer schemas don't need to be here because client doesn't know about those but since they are here for historical reasons we keep it that way for consistency.

PR #169, opened 2024-04-16
Migrate Buildkite CI queues from AWS to GKE

mstifflin:
**What changed?**
- Update Buildkite pipeline yaml to work with the newly provisioned queues in Google Kubernetes Engine
- Use [agent-stack-k8s v0.8.0 helm chart](https://github.com/buildkite/agent-stack-k8s/tree/v0.8.0?tab=readme-ov-file#cloning-repos-via-ssh) which has its own expected pipeline yaml syntax in order to successfully onboard. 

**Why?**
- There is a company mandate to evacuate Uber's AWS CI footprint. The go forward decision is to use Google Cloud. Buildkite Enterprise recommends GKE (as opposed to GCP) as the way to have a queue with autoscaling compute. 

**How did you test it?**
- https://buildkite.com/uberopensource/cadence-idl/builds/518

**Potential risks**
- CI builds will be broken or flaky
- Can be mitigated by a git revert

**Release notes**
n/a

**Documentation Changes**
n/a

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,16 +1,42 @@
+container:
+  kubernetes: &kubernetes
+    gitEnvFrom:
+      - secretRef:
+          name: oss-github-ssh-credentials
+    podSpec: &podSpec
+      containers:
+        - &commandContainer
+          image: us-west1-docker.pkg.dev/ci-compute/buildkite-images/buildkite-command-container:v2
+          command:
+          - |-
+            echo "Command step was not overridden."
+            exit 1
+
 steps:
   - label: "fossa analyze"
     agents:
       queue: "buildkite-gcp"
-    command: "./scripts/buildkite/fossa.sh"
+    plugins:
+      - kubernetes:
+          <<: *kubernetes
+          podSpec:
+            <<: *podSpec
+            containers:
+              - <<: *commandContainer
+                command:
+                - |-
+                  ./scripts/buildkite/fossa.sh
+
   - label: "proto lint"
     agents:
       queue: "buildkite-gcp"
     plugins:
       - kubernetes:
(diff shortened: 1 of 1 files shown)


taylanisikdemir (approved):

Groxx:
bleh.  that's quite a lot of `<<: *...` noise...

... and is likely as non-repeated as is possible here, so 👍.  just kinda lamenting how noisy this kind of yaml stuff grows to be, especially since this will probably be repeated a few more times for the client and server repos.

I don't suppose there are less array-blended versions of this config, which might be easier to read and maintain?  The details of how yaml merges this are, in my experience, not very well understood by most people.

mstifflin:
I agree, the yaml merges are difficult to read. I had to use `yq 'explode(.)' .buildkite/pipeline.yml` for my own sanity checking and debugging when I originally wrote it, which I could add as a comment to the top of the file for anyone in the future who wants to modify it. 

> I don't suppose there are less array-blended versions of this config, which might be easier to read and maintain?
Unfortunately, we have a list of constraints:
1. We must use auto scaling machines rather than having compute available at all times
2. The Buildkite recommended way to have autoscaling agents in GCP is k8s + agent-stack-k8s
3. agent-stack-k8s is still in early development (we're using v0.8.0) and for whatever reason this verbosity is how they decided to implement it
4. There is a timeline expectation for evacuating AWS.

BK Enterprise released v0.9.0 a couple weeks ago then v0.10.0 quickly after for bug fixes, so I'm holding off onboarding onto those newer versions until they're more established. 

Finally, Buildkite Enterprise recently released a "managed infra" product that we are considering using for OSS pipelines, which would allow us to move away from agent-stack-k8s. TBD if whoever is in charge of making those decisions will actually decide to pay for it though.

PR #173, opened 2024-07-09
Introduce ListAllWorkflowExecutions Api

sankari165:
### Summary
New api to list all workflow executions with support for customised sorting, partial search and filtering on multiple statuses

### Type of Change
- [x] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect
No

### Detailed Description

### Impact Analysis
Unused Api

### Testing Plan
- **Unit Tests**: [Do we have unit test covering the change?] Yes
- **Persistence Tests**: [If the change is related to a data type which is persisted, do we have persistence tests covering the change?] N/A
- **Integration Tests**: [Do we have integration test covering the change?] No
- **Compatibility Tests**: [Have we done tests to test the backward and forward compatibility?] N/A

### Rollout Plan
- What is the rollout plan? Make the api available in server before making it available to clients
- Does the order of deployment matter? The api should be available in server for usage by the client
- Is it safe to rollback? Does the order of rollback matter? Safe to rollback as it is not used anywhere


proto/uber/cadence/api/v1/service_visibility.proto
```diff
@@ -66,6 +69,23 @@ message ListWorkflowExecutionsResponse {
   bytes next_page_token = 2;
 }
 
+message ListAllWorkflowExecutionsRequest {
+  string domain = 1;
+  int32 maximum_page_size = 2;
```
taylanisikdemir, line 74: this is called `page_size` in `ListWorkflowExecutionsRequest` so let's stick with it for consistency

proto/uber/cadence/api/v1/service_visibility.proto
```diff
@@ -66,6 +69,23 @@ message ListWorkflowExecutionsResponse {
...
+  bytes next_page_token = 3;
+  StartTimeFilter start_time_filter = 4;
+  bool partial_match = 5;
+  repeated StatusFilter close_status = 6;
+  string workflow_search_value = 7;
```
taylanisikdemir, line 79: what is this field? would be nice to add one line comment on each of these fields in the request

thrift/shared.thrift
```diff
@@ -1456,6 +1456,23 @@ struct ListWorkflowExecutionsResponse {
   20: optional binary nextPageToken
 }
 
+struct ListAllWorkflowExecutionsRequest {
+  10: optional string domain
+  20: optional i32 maximumPageSize
```
taylanisikdemir, line 1461: same here

taylanisikdemir (approved):

PR #174, opened 2024-07-12
Update changes to listallworkflowexecutions api

sankari165:
### Summary
Minor changes to field names were not updated in the generated files in #173

### Type of Change
- [x] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect
- [ ] Does it change the data stored in database? No

### Detailed Description

### Impact Analysis
Unused Api

### Testing Plan
This is follow up to #173

### Rollout Plan
This is follow up to #173
- What is the rollout plan?


3vilhamster (approved):
nit: we can create a workflow to ensure that files are updated.

PR #175, opened 2024-07-19
Add an optional First_Run_At field to StartWorkflowOptions

timl3136:
### Summary
We are adding another layer of customization which allows users to define a specific time to start the run of their workflow. (In case of cron workflow, it would be only applied to the first run of the workflow.)

The new first_run_at fields will be applied and override other delay start related parameters (delay_start_seconds, jitter_start_seconds, cron_schedule) under the following conditions:
First_run_at > Current time

This condition guarantee two things:
The logic is only triggered when the user specifies a first run timestamp
AND that timestamp is only triggered ONCE hence not interfering with other scheduling logic in later runs

### Type of Change
- [x] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect
- [ ] Does it change the data stored in database? No

### Detailed Description
Add a new start workflow option called First_Run_At, which would affect the scheduled time for first run of a workflow.

### Impact Analysis
- It’s an additional optional field so it should be compatible with all clients as the processing happens in the server
- In terms of server backward compatibility, old server should be compatible with new requests as the field is added to the end of the request and will be dropped. The logic when this field is ignored will be the same as the older version.

- It’s forward compatible because new server will treat the value as unset when the old request that's missing that field is received.

### Testing Plan
- Yes, it would be implemented in both server and client repo
- The data will not be persisted

### Rollout Plan
- What is the rollout plan?
1. Add and merge the fields in cadence-idl repo
2. Update SHA for cadence-idl repo and handled data fields in shared.go in cadence-server repo
3. Rollout to all clusters before updating SHA on cadence-client
4. Updating SHA for cadence-idl and add the fields in cadence-client
6. Update client version in cadence-canary for monitoring
7. Add canary workflows to ensure the new field work as expected
8. Release new version in cadence-client with the added field

Yes, we need to get the server build updated before deploying such change in client.
Yes, rollback server version is safe given that the change is backward compatible.
No


Groxx:
Just tossing some thoughts while thinking through this:
- client library probably doesn't care at all about this field, so:
  - new clients starting + old clients working => field is silently ignored which is fine, clients don't decide when things happen.
  - old clients starting + new clients working => field is not set and empty values are ignred, which is fine.
- [ ] new RPC request to old servers: silently dropped, possibly an issue (but only during rollout and rollback)
  - is our plan to just have this rolled out in server well before client-side use, or should we try to do something to reject newer requests that use the field?
- old request to new servers: seems fine :thumbsup:


a bit more questionable (though I realize this is just rpc, not storage, but it implies storage of some kind...):
- [ ] new request to new servers -> rollback: what will this do to task processing, replication, etc?
  - how will old server instances behave when this field is silently ignored?
  - do we need to consider this a "do not roll back past this point" marker?

timl3136:
During the rollout, we will roll out the change to all the servers completely before rollout to client.
In terms of rollback the server to older version, the compatibility should be just fine as the field would be dropped by the old server version.

taylanisikdemir (approved):

PR #179, opened 2024-08-19
Add Tasklist not owned by host error, so we can propagate it internaally between matching and history 

jakobht:
### Summary
Add a new error type so we can send the error from matching to history, such that history can handle the error more gracefully
This PR uses the change: https://github.com/uber/cadence/pull/6233 

### Type of Change
- [x] Add new data type(s)
- [ ] Add new API(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

No

### Detailed Description
- Added new exception in the thrift file. The corresponding error is added in proto in the main repo, as the is purely for internal usage. The exception's definition is: 
```
exception TaskListNotOwnedByHostError {
    1: required string ownedByIdentity
    2: required string myIdentity
    3: required string tasklistName
}
```

- Added the exception as returned by all the matching API functions. 

### Impact Analysis

All internal endpoints can handle unknown errors, so compatibility should not be an issue. 

### Testing Plan
We have unit tests in the main repo (see above) 

Not diretly 

The change has been deployed in staging without any issues.

### Rollout Plan
- What is the rollout plan?
The change will be deployed using CD, the feature flag `matching.enableTasklistGuardAgainstOwnershipLoss` will be used to gradually enable the error returning.
No
Yes it is safe, the order does not matter
Yes there is a dynamic config flag `matching.enableTasklistGuardAgainstOwnershipLoss` that will stop matching returning this error.

### DIFF
--- a/thrift/matching.thrift
+++ b/thrift/matching.thrift
@@ -149,6 +149,7 @@ service MatchingService {
       2: shared.InternalServiceError internalServiceError,
       3: shared.LimitExceededError limitExceededError,
       4: shared.ServiceBusyError serviceBusyError,
+      5: shared.TaskListNotOwnedByHostError taskListNotOwnedByHostError,
     )
 
   /**
@@ -161,6 +162,7 @@ service MatchingService {
       2: shared.InternalServiceError internalServiceError,
       3: shared.LimitExceededError limitExceededError,
       4: shared.ServiceBusyError serviceBusyError,
+      5: shared.TaskListNotOwnedByHostError taskListNotOwnedByHostError,
     )
 
   /**
@@ -176,6 +178,7 @@ service MatchingService {
       5: shared.DomainNotActiveError domainNotActiveError,
       6: shared.RemoteSyncMatchedError remoteSyncMatchedError,
       7: shared.StickyWorkerUnavailableError stickyWorkerUnavailableError,
+      8: shared.TaskListNotOwnedByHostError taskListNotOwnedByHostError,
     )
 
   /**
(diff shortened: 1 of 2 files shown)


thrift/matching.thrift
```diff
@@ -149,6 +149,7 @@ service MatchingService {
       2: shared.InternalServiceError internalServiceError,
       3: shared.LimitExceededError limitExceededError,
       4: shared.ServiceBusyError serviceBusyError,
+      5: shared.TaskListNotOwnedByHostError taskListNotOwnedByHostError,
```
taylanisikdemir, line 152: what happens to the old clients (workers that is deployed before this change) when they receive this error? 
For some reason worker code is very strict about responses and can cause crashes for unrecognized enums. Not sure what the behavior is for unrecognized exceptions. It's worth to try

jakobht: The error will not be returned from the Poll requests. An empty response will be returned instead (updated here and in the PR in the server repo). 

In general frontend will convert this error to a generic system error as frontend does not know about it. If this error after retrys etc is still returned, it should be considered a generic system error, so that is what we want.

dkrotx (approved):

davidporter-id-au (approved):
nice!

thrift/shared.thrift
```diff
@@ -114,6 +114,12 @@ exception StickyWorkerUnavailableError {
   1: required string message
 }
 
+exception TaskListNotOwnedByHostError {
```
davidporter-id-au, line 117: I wondered if it would be worth having a `NotOwnedByHostError` for use in both history and matching and while I can see that being slightly less annoying, it may not be great ultimately to tie the two errors together, so I ultimately don't really have a strong opinion on this. It may be easier to evolve them separately.

davidporter-id-au:
Should there be a PR for GRPC too?

PR #183, opened 2024-09-04
[Wf-diagnostics] Introduce a new api to diagnose a workflow execution

sankari165:
### Summary
[Brief summary of the change, including the rationale and intended effect] This api will trigger the start of Diagnostics workflow that runs in the cadence worker service. Having this as a api in frontend, allows the use of workflow diagnostics from cli or web 

### Type of Change
- [x] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect
- [ ] Does it change the data stored in database? N/A

### Detailed Description
Introduction of new api to trigger workflow to perform diagnostics from cadence frontend. The result of the diagnostics will be returned in a json format. This api will be triggered from cli and web to provide diagnostics for a workflow execution to the users.

### Impact Analysis
- **Backward Compatibility**: [Analysis of backward compatibility] N/A
- **Forward Compatibility**: [Analysis of forward compatibility] N/A

### Testing Plan
No compatibility issues as it is a new unused API

### Rollout Plan
- What is the rollout plan? Rollout to all cadence clusters 
- Does the order of deployment matter? no
- Is it safe to rollback? Does the order of rollback matter? no issues as it is a new unused API


proto/uber/cadence/api/v1/service_workflow.proto
```diff
@@ -108,6 +111,16 @@ message RestartWorkflowExecutionRequest {
...
 }
 
+message DiagnoseWorkflowExecutionRequest {
+  string domain = 1;
+  WorkflowExecution workflow_execution = 2;
+  string identity = 3;
+}
+
+message DiagnoseWorkflowExecutionResponse {
+  Payload response = 1;
```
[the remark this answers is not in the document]
sankari165: updated the response

proto/uber/cadence/api/v1/service_workflow.proto
```diff
@@ -108,6 +111,17 @@ message RestartWorkflowExecutionRequest {
...
 
+message DiagnoseWorkflowExecutionRequest {
+  string domain = 1;
+  WorkflowExecution workflow_execution = 2;
+  string identity = 3;
+}
+
+message DiagnoseWorkflowExecutionResponse {
+  string domain = 1;
+  WorkflowExecution workflow_execution = 2;
```
3vilhamster, line 122: nit a better name to be explicit on what is in the response workfow.
```suggestion
  WorkflowExecution diagnostic_workflow_execution = 2;
```

3vilhamster (approved):
nit for a better naming, but overall it is good to land.

proto/uber/cadence/api/v1/service_workflow.proto
sankari165: updated

PR #189, opened 2024-11-17
added domainId to RegisterDomainRequest

samkitshah1262:
### Summary
Towards [Issue](https://github.com/uber/cadence/issues/5350)

### Type of Change
- [ ] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description
Added field domainId in the requests to support the functionality of using user provided domain id at the time of domain creation.

### Impact Analysis
- **Backward Compatibility**: [Analysis of backward compatibility] Backwards compatible

### Testing Plan
- **Unit Tests**: [Do we have unit test covering the change?] Yes
- **Persistence Tests**: [If the change is related to a data type which is persisted, do we have persistence tests covering the change?] Yes

### Rollout Plan
- What is the rollout plan?

### DIFF
--- a/proto/uber/cadence/api/v1/service_domain.proto
+++ b/proto/uber/cadence/api/v1/service_domain.proto
@@ -67,6 +67,7 @@ message RegisterDomainRequest {
   string history_archival_uri = 11;
   ArchivalStatus visibility_archival_status = 12;
   string visibility_archival_uri = 13;
+  string domain_id = 14;
 }
 
 message RegisterDomainResponse {
--- a/proto/uber/cadence/api/v1/service_domain.proto
+++ b/proto/uber/cadence/api/v1/service_domain.proto
@@ -54,6 +54,9 @@ service DomainAPI {
 }
 
 message RegisterDomainRequest {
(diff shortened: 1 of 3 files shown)


proto/uber/cadence/api/v1/service_domain.proto
```diff
@@ -67,6 +67,7 @@ message RegisterDomainRequest {
   string history_archival_uri = 11;
   ArchivalStatus visibility_archival_status = 12;
   string visibility_archival_uri = 13;
+  string domain_id = 14;
```
taylanisikdemir, line 70: please add a comment on this field
```
// domain_id is optional. A random UUID is generated if not set. 
// It can be set to a custom UUID as long as it doesn't collide with id of other existing domains
// If unsure leave this empty.
string domain_id = 14;
```

taylanisikdemir:
The corresponding issue is closed as it turns out this is not a need. Closing the PR. 
Feel free to pick up other issues if you would like to contribute. Also check with the team on slack channel if you are unsure.

PR #193, opened 2024-12-26
Add TaskList Isolation Group Assignment

natemort:
### Summary
Add fields to associate individual TaskList partitions with isolation groups and collect metrics from each TaskList partition at the isolation group granularity.

### Type of Change
- [ ] Add new API(s)
- [x] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description
Update TaskListPartitionConfig with an explicit mapping of partition ids to metadata about the partition, rather than just a number of partitions. Store the isolation groups associated with each partition there.

Update TaskListStatus to include metrics about the TaskList with a per isolation group granularity. This is returned as part of DescribeTaskList, which will be used by the root partition to aggregate metrics from each of its children. It's currently used only to check if a child partition has finished draining.

The TaskListPartitionConfig type is not stored directly in Cassandra, and is instead represented by a UDT. We will do the same for TaskListPartition.

### Impact Analysis
- **Backward Compatibility**: So long as we continue to populate both the num_* fields we can also include the partitions field, so long as we ensure that the partition ids are contiguous.

### Testing Plan
As we are only adding new fields we have no current tests covering this change. This will be covered by unit, persistence, and integration tests once it is merged 

### Rollout Plan
- What is the rollout plan? Ideally we deploy this (and related changes) alongside the changes migrating from dynamic config to Cassandra for the sake of simplicity. It's safe either way.
- Does the order of deployment matter? No
- Is it safe to rollback? Does the order of rollback matter? Rolling back is safe
- Is there a kill switch to mitigate the impact immediately? Disabling reading the config from Cassandra.


natemort:
Recreating with correct branch, accidentally pushed my branch to the repo.

PR #194, opened 2024-12-27
Add TaskList Isolation Group Assignment

natemort:
### Summary
Add fields to associate individual TaskList partitions with isolation groups and collect metrics from each TaskList partition at the isolation group granularity.

### Type of Change
- [ ] Add new API(s)
- [x] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description
Update TaskListPartitionConfig with an explicit mapping of partition ids to metadata about the partition, rather than just a number of partitions. Store the isolation groups associated with each partition there.

Update TaskListStatus to include metrics about the TaskList with a per isolation group granularity. This is returned as part of DescribeTaskList, which will be used by the root partition to aggregate metrics from each of its children. It's currently used only to check if a child partition has finished draining.

The TaskListPartitionConfig type is not stored directly in Cassandra, and is instead represented by a UDT. We will do the same for TaskListPartition.

### Impact Analysis
- **Backward Compatibility**: So long as we continue to populate both the num_* fields we can also include the partitions field, so long as we ensure that the partition ids are contiguous.

### Testing Plan
As we are only adding new fields we have no current tests covering this change. This will be covered by unit, persistence, and integration tests once it is merged 

### Rollout Plan
- What is the rollout plan? Ideally we deploy this (and related changes) alongside the changes migrating from dynamic config to Cassandra for the sake of simplicity. It's safe either way.
- Does the order of deployment matter? No
- Is it safe to rollback? Does the order of rollback matter? Rolling back is safe
- Is there a kill switch to mitigate the impact immediately? Disabling reading the config from Cassandra.


proto/uber/cadence/api/v1/tasklist.proto
```diff
@@ -57,12 +57,24 @@ message TaskListPartitionMetadata {
...
 message TaskListStatus {
   int64 backlog_count_hint = 1;
   int64 read_level = 2;
   int64 ack_level = 3;
+  // The rate at which tasks may be dispatched according to poller configuration
   double rate_per_second = 4;
   TaskIDBlock task_id_block = 5;
+  map<string, IsolationGroupMetrics> isolation_group_metrics = 6;
+  // The rate at which tasks are being added to this partition
+  double new_tasks_per_second = 7;
```
timl3136, line 77: I notice the isolation_group_metrics also has a new_tasks_per_second field, are those two field's going to have the same value or different value?
- If they are going to be the same, do we necessary needs two fields?
- If different, maybe we can rename the new_tasks_per_second in isolationGroupMetrics to be "isolation_group_new_tasks_per_second"?

Additionally, the existing rate_per_second might be confused with the new_tasks_per_second. In my understanding, the new_tasks_per_second might override the existing rate_per_second? We might need more clarification there.

taylanisikdemir (approved):

proto/uber/cadence/api/v1/tasklist.proto
natemort: I've updated the comments to better clarify the intention. The two `new_tasks_per_second` fields represent the same value but in different contexts: In `IsolationGroupMetrics` it's the value for an individual isolation group, in `TaskListStatus` it represents the entire partition of the task list. I think them having the same name is reasonable since they're conceptually the same value. It's similar to how both a `Circle` and `Square` might have a field called `area`.

Agreed that the existing `rate_per_second` field is confusing. I've added comments to both fields to better distinguish them. They represent entirely different values. The `rate_per_second` is a rate limit on how quickly the task list can dispatch tasks. Users can set it in their worker configuration, which the poller sends to the server with each request.

proto/uber/cadence/api/v1/tasklist.proto
```diff
@@ -81,8 +93,14 @@ message StickyExecutionAttributes {
...
+  repeated string isolation_groups = 1;
+}
+
 message TaskListPartitionConfig {
   int64 version = 1;
-  int32 num_read_partitions = 2;
-  int32 num_write_partitions = 3;
+  int32 num_read_partitions = 2 [deprecated = true];
+  int32 num_write_partitions = 3 [deprecated = true];
+  map<int32, TaskListPartition> read_partitions = 4;
```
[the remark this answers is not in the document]
natemort: Done.

timl3136 (approved):

PR #202, opened 2025-05-13
Add active clusters field to domain for active-active mode

taylanisikdemir:
### Summary

### Type of Change
- [ ] Add new API(s)
- [x] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description

Defining ActiveClusters struct and adding it to relevant domain objects in APIs and SQL blobs. See [design doc](https://github.com/cadence-workflow/cadence/blob/master/docs/design/active-active/active-active.md) for more details.  It should be a low risk change until domains in prod start switching to active-active mode. 

### Impact Analysis

### Testing Plan
   Yes and will add more while pulling this into server repo
    Yes
   Yes and will add more while pulling this into server repo
    WIP

### Rollout Plan
- What is the rollout plan?
Plan is to merge the changes into server repo and get it deployed to all services across all envs before onboarding any user to this feature. Until then it should be no-op.

No

Safe to rollback until users start onboarding to active-active domains.

Rollback is the kill switch.


thrift/shared.thrift
```diff
@@ -1087,6 +1108,8 @@ struct RegisterDomainRequest {
   50: optional bool emitMetric = true
   60: optional list<ClusterReplicationConfiguration> clusters
   70: optional string activeClusterName
+  // activeClusters is a map of region name to active cluster name for active-active domain
+  75: optional map<string, string> activeClustersByRegion
```
taylanisikdemir, line 1112: This is <string, string> map because all we need is cluster name per region during domain registration. Initial failover versions will be assigned by domain handler.

thrift/shared.thrift
```diff
@@ -50,6 +50,8 @@ exception EntityNotExistsError {
   1: required string message
   2: optional string currentCluster
   3: optional string activeCluster
+  // activeClusters is a list of active clusters for active-active domain
+  4: required list<string> activeClusters
```
Shaddoll, line 54: missing corresponding changes in proto

thrift/shared.thrift
Shaddoll: It's different from proto, which makes it complicated to create mappers.

thrift/shared.thrift
```diff
@@ -1075,8 +1079,25 @@ struct ClusterReplicationConfiguration {
...
+  10: optional map<string, ActiveClusterInfo> activeClustersByRegion
+}
+
+// ActiveClusterInfo contains the configuration of active-active domain's active cluster & failover version for a specific region
+struct ActiveClusterInfo {
+  10: optional string activeClusterName
+  20: optional i64 (js.type = "Long") failoverVersion
 }
 
 struct RegisterDomainRequest {
```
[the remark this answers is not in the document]
taylanisikdemir: `UpdateDomainRequest` has `DomainReplicationConfiguration` field which contains the new field

thrift/shared.thrift
taylanisikdemir: updated proto side to match this.

Shaddoll (approved):

PR #205, opened 2025-06-03
Add a new bool flag of BufferOneCronWorkflow to startWorkflowOptions

timl3136:
### Summary

### Type of Change
- [x] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description

### Impact Analysis

### Testing Plan
- Yes, it would be implemented in both server and client repo
- The data will not be persisted
- We will add new integration tests on cron workflows

### Rollout Plan
- What is the rollout plan?
1. Add and merge the fields in cadence-idl repo
2. Update SHA for cadence-idl repo and handled data fields in shared.go in cadence-server repo
3. Rollout to all clusters before updating SHA on cadence-client
4. Updating SHA for cadence-idl and add the fields in cadence-client
5. Update client version in cadence-canary for monitoring
6. Add canary workflows to ensure the new field work as expected
7. Release new version in cadence-client with the added field
Yes, we need to get the server build updated before deploying such change in client.
Yes, rollback server version is safe given that the change is backward compatible.


thrift/shared.thrift
```diff
@@ -1192,6 +1196,7 @@ struct StartWorkflowExecutionRequest {
   160: optional i32 delayStartSeconds
   170: optional i32 jitterStartSeconds
   180: optional i64 (js.type = "Long") firstRunAtTimestamp
+  190: optional bool bufferOneCronWorkflow
```
davidporter-id-au, line 1199: uh.. @timl3136  sorry to be annoying,  but bools in IDL are usually a bit of a worry because they only represent two things, but usually there's more than 2 eventually. Is there ever a case that we may want to buffer N cron workflows (I have no idea what this does)

timl3136: In that case, I would like to rename the field to cron_overlap_policy since bufferOneCronWorkflow only reflects one type of behavior.

davidporter-id-au (approved):
ty!

thrift/shared.thrift
```diff
@@ -324,6 +324,11 @@ enum ArchivalStatus {
   ENABLED,
 }
 
+enum CronOverlapPolicy {
+  DISABLED,
```
taylanisikdemir, line 328: SKIP instead of DISABLED would be better naming

proto/uber/cadence/api/v1/workflow.proto
```diff
@@ -153,6 +153,12 @@ enum WorkflowIdReusePolicy {
   WORKFLOW_ID_REUSE_POLICY_TERMINATE_IF_RUNNING = 4;
 }
 
+enum CronOverlapPolicy {
+  CRON_OVERLAP_POLICY_INVALID = 0;
+  CRON_OVERLAP_POLICY_DISABLED = 1;
```
taylanisikdemir, line 158: same here

taylanisikdemir (approved):

thrift/sqlblobs.thrift
```diff
@@ -147,6 +147,7 @@ struct WorkflowExecutionInfo {
   128: optional map<string, string> partitionConfig
   130: optional binary checksum
   132: optional string checksumEncoding
+  134: optional CronOverlapPolicy cronOverlapPolicy
```
taylanisikdemir, line 150: this should be `shared.CronOverlapPolicy`

PR #206, opened 2025-06-04
[active-active] Add active cluster selection policy to workflow start

taylanisikdemir:
### Summary
Active cluster selection policy will be applicable to active-active domains. More details on active-active mode can be found [here](https://github.com/cadence-workflow/cadence/blob/master/docs/design/active-active/active-active.md).

### Type of Change
- [x] Add new API(s)
- [x] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description
There will be a new field in workflow execution info. This is mostly for informational purposes.
There is going to be a new row type to store the policy for wf-runid pair which will be used by frontend to determine active region/cluster to forward accordingly.

### Impact Analysis
Changes are backward compatible. No impact until server starts using these.

### Testing Plan

### Rollout Plan
- What is the rollout plan?


davidporter-id-au (approved):

thrift/shared.thrift
```diff
@@ -2101,3 +2107,22 @@ struct TaskKey {
...
+  20: optional string stickyRegion
+
+  // external_entity_type/external_entity_key is the type/key of the external entity if strategy is ACTIVE_CLUSTER_SELECTION_STRATEGY_EXTERNAL_ENTITY
+  // external entity type must be one of the supported types in active cluster manager. Custom ones can be added by implementing the corresponding interface.
+  30: optional string externalEntityType
+  40: optional string externalEntityKey
+}
+
+enum ActiveClusterSelectionStrategy {
+  Invalid,
```
timl3136, line 2125: thrift does not use a invalid enum (unlike proto), also, it might be better to put them along with other enums around L339.
It also should be all cap letters

timl3136 (changes requested):
I think we should fix the enum naming before merging

proto/uber/cadence/api/v1/common.proto
```diff
@@ -135,3 +135,22 @@ message AsyncWorkflowConfiguration {
...
+  ACTIVE_CLUSTER_SELECTION_STRATEGY_REGION_STICKY = 1;
+  ACTIVE_CLUSTER_SELECTION_STRATEGY_EXTERNAL_ENTITY = 2;
+}
+
+message ActiveClusterSelectionPolicy {
+  ActiveClusterSelectionStrategy strategy = 1;
+
+  // sticky_region is the region sticky if strategy is ACTIVE_CLUSTER_SELECTION_STRATEGY_REGION_STICKY
+  // This is the default strategy for active-active domains and region would be set to receiver cluster's region if not specified.
+  string sticky_region = 2;
```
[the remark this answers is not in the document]
taylanisikdemir: I didn't use it because we also have corresponding thrift structs and it doesn't support oneof. However, this hasn't stopped us using oneof in proto version of those structs so let me use it to be consistent with existing code even though it's not going to be consistent between proto & thrift :)

thrift/shared.thrift
taylanisikdemir: thanks fixed

PR #209, opened 2025-06-10
Add TaskListKind Ephemeral

natemort:
### Summary
Add support for Ephemeral TaskLists, which will automatically be deleted once no longer in use. This is behavior already supported by Sticky TaskLists, but non-sticky TaskList scenarios such as sessions and integration tests both require generating dynamic TaskLists. We need similar functionality for these TaskLists.

Requests to Start/SignalWithStart a workflow specify a TaskList instance (Name + Kind), so no changes are needed for those.

### Type of Change
- [ ] Add new API(s)
- [x] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description

Add new TaskListKind EPHEMERAL.

Add a TaskList to the response of DescribeTaskList so that the server can expose the TaskListKind of a given TaskList. The request currently includes a TaskList, but the TaskListKind is not considered part of the identity of a TaskList and so the Kind field there is ignored. As a result there isn't a way to determine the Kind of a TaskList beyond inferring it from the name or looking at the database.

Add a TaskListKind field to WorkflowExecutionInfo. This allows for DescribeWorkflow to return the TaskListKind the workflow was started with. Currently this is implicitly TaskListKind NORMAL.

Update sqlblobs WorkflowExecutionInfo so that we can store it for SQL persistence types.

### Impact Analysis
- **Backward Compatibility**: Adding new enum types is not backward compatible. Once we have written data with this new type we are unable to rollback. Additionally we need to ensure that clients do not receive this new value before they understand it, as they otherwise will panic. Sessions and our uber-internal integration tests are both host-specific tasklists, which significantly mitigates the risk of this change. The client needs to Start a workflow (or start polling the session tasklist) with TaskListKind EPHEMERAL before it would ever be returned.
- **Forward Compatibility**: These changes add support for any additional TaskListKinds we may add in the future.

### Testing Plan
- **Unit Tests**: Will be added to Cadence
- **Persistence Tests**:  Will be added to Cadence
- **Integration Tests**: Will be added to Cadence
- **Compatibility Tests**: Will be added to Cadence.

### Rollout Plan
- What is the rollout plan? 
  - Deploy IDL/server changes, along with a client version containing the IDL changes. 
  - Wait for it to be deployed fully.
  - Then implement support within the client (behind a flag) to use these values.
  - Not between Cadence service types, but it does matter between the client and server.
  - Once the server has written workflows with this new type we cannot roll back.
  - No


proto/uber/cadence/api/v1/workflow.proto
```diff
@@ -45,10 +45,12 @@ message WorkflowExecutionInfo {
   SearchAttributes search_attributes = 10;
   ResetPoints auto_reset_points = 11;
   string task_list = 12;
+  TaskListKind task_list_kind = 17;
```
Shaddoll, line 48: nit: How about introducing `TaskList task_list_metadata` and marking `string task_list` deprecated?

Shaddoll (approved):

proto/uber/cadence/api/v1/workflow.proto
natemort: Switched to `task_list_info`

taylanisikdemir (approved):

PR #215, opened 2025-07-23
Introduce predicate to virtual slice state to support task filtering

Shaddoll:
### Summary

### Type of Change
- [ ] Add new API(s)
- [x] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description

### Impact Analysis

### Testing Plan

### Rollout Plan
- What is the rollout plan?

### DIFF
--- a/thrift/shared.thrift
+++ b/thrift/shared.thrift
@@ -2098,6 +2098,7 @@ struct VirtualQueueState {
 
 struct VirtualSliceState {
   10: optional TaskRange taskRange
+  20: optional Predicate predicate
 }
 
 struct TaskRange {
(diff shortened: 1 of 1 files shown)


thrift/shared.thrift
```diff
@@ -2127,3 +2128,25 @@ enum ActiveClusterSelectionStrategy {
...
+struct DomainIDPredicateAttributes {
+  10: optional list<string> domainIDs
+  20: optional bool isExclusive
+}
+
+struct Predicate {
+  10: optional PredicateType predicateType
+  20: optional UniversalPredicateAttributes universalPredicateAttributes
+  30: optional EmptyPredicateAttributes emptyPredicateAttributes
+  70: optional DomainIDPredicateAttributes domainIDPredicateAttributes
```
taylanisikdemir, line 2151: any reason to jump to 70?

taylanisikdemir (approved):

PR #222, opened 2025-09-30
refactor: Replace region concept with ClusterAttribute in active-active domains

davidporter-id-au:
### Summary

This is is a semi-breaking change of a prerelease feature which is being refactored. The only reason for the breaking change is that it's not fully released yet and therefore is unlikely to have any dependencies on it.

### Type of Change
- [ ] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)
- [x] Other (please provide detailed description)

### Data Effect

### Detailed Description

### Impact Analysis

### Testing Plan

### Rollout Plan
- What is the rollout plan?

### DIFF
--- a/thrift/shared.thrift
+++ b/thrift/shared.thrift
@@ -51,8 +51,6 @@ exception EntityNotExistsError {
   2: optional string currentCluster
   3: optional string activeCluster
   4: required list<string> activeClusters // todo(david.porter) remove as its disused
-  // activeClusters is a list of active clusters for active-active domain
-  5: required ActiveClusters activeClusters
 }
 
 exception ServiceBusyError {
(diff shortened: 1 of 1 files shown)


thrift/shared.thrift
```diff
@@ -1127,8 +1132,8 @@ struct RegisterDomainRequest {
   50: optional bool emitMetric = true
   60: optional list<ClusterReplicationConfiguration> clusters
   70: optional string activeClusterName
-  // activeClusters is a map of region name to active cluster name for active-active domain
-  75: optional map<string, string> activeClustersByRegion
+  // activeClusters is a map of cluster-attribute name to active cluster name for active-active domain
+  75: optional map<string, string> activeClustersByClusterAttribute
```
Shaddoll, line 1136: this doesn't look correct to me

Shaddoll (approved):

PR #224, opened 2025-10-01
feat: adding field for active/active use-case

davidporter-id-au:
### Summary

This adds active/active cluster-attributes to the request payload of the new 'Failover' endpoint for the intent of allowing it to work with this kind of domain. 

It also tidies up two other fields that were inconsistent with a breaking-but-not-yet-used-field name change.

### Type of Change
- [x] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description

### Impact Analysis

### Testing Plan

### Rollout Plan
- What is the rollout plan?


proto/uber/cadence/api/v1/service_domain.proto
```diff
@@ -78,7 +78,7 @@ message RegisterDomainRequest {
   ArchivalStatus visibility_archival_status = 12;
   string visibility_archival_uri = 13;
   map<string, string> active_clusters_by_region = 14; // todo (david.porter) to remove as this is no longer used
-  ActiveClusters active_clusters_by_cluster_attribute = 15;
```
davidporter-id-au, line 81: This is a breaking change, but this feature's not implemented yet, so it shouldn't be a concern. 

The actual field has - inside it - fields by cluster-attributes, so adding it externally here as a name is redundant

PR #259, opened 2026-04-14
revert: Upgrade YARPC package

c-warren:
Reverts cadence-workflow/cadence-idl#255


shijiesheng (approved):

c-warren:
This downgrade was to remove the yarpc upgrade that in turn upgrades grpc within our cadence server, which in turn is not compatible with some of our internal systems. However using yarpc 1.88 is fine with a compatible version of grpc, so we can leave this as is.

PR #260, opened 2026-04-22
feat: Add base_name field to TaskList

Shaddoll:
### Summary
Add base_name field to TaskList type

### Type of Change
- [ ] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description
Add base_name field to TaskList type

### Impact Analysis
- **Backward Compatibility**: Yes.
- **Forward Compatibility**: Yes.

### Testing Plan
- **Unit Tests**: Will add in server repo.
- **Persistence Tests**: N/A
- **Integration Tests**: N/A
- **Compatibility Tests**: N/A

### Rollout Plan
- What is the rollout plan? 
- Does the order of deployment matter? No.
- Is it safe to rollback? Does the order of rollback matter? Yes. No.
- Is there a kill switch to mitigate the impact immediately? No.

----
## Summary by Gitar

- **Schema updates:**
  - Added `base_name` string field to `TaskList` in both Protobuf (`tasklist.proto`) and Thrift (`shared.thrift`) definitions.
  - Updated generated Go code for the `TaskList` struct to include the new `base_name` field.

<sub>This will update automatically on new commits.</sub>


timl3136 (approved):
nit question: should we put comment describing base_name in tasklist.proto? it will be hard to find this in the future

proto/uber/cadence/api/v1/tasklist.proto
```diff
@@ -34,6 +34,17 @@ import "google/protobuf/wrappers.proto";
...
+  //
+  // - For STICKY task lists, this is the associated normal task list.
+  // - For EPHEMERAL task lists, this is the associated normal task list.
+  // Ephemeral task lists are typically used for test isolation.
+  // - For NORMAL task lists, this field is usually empty, but may be set
+  // in certain cases (e.g., Go session worker activity task lists).
+  //
+  // This field is used as an identifier for rate limiting and task prioritization
+  // in the history scheduler.
+  string base_name = 3;
```
c-warren, line 47: We chatted about this internally but I'd like to put it here for discussion in the future as well. 

I think there is great value in making the `name` field immutable. It will ensure that whenever we are accessing the TaskList, we are guaranteed that it is the original task list name specified by the user. This will simplify any logic that is accessing the task list - we don't need to manage whether or not it is the sticky, ephemeral, or original name with helper functions, when thinking about task lists we don't have to think of multiple different fields to determine what the name may be. 

Having a computed_name for the name after it has been mutated by the sdk/someone else makes the behavior of the task list more clear as well. This would also make the field more extensible in the future, without having to complicate it further. 

I know migrating to this model would be hard, but it would (eventually) simplify our logic and simplify thinking about task list features.

c-warren:
> nit question: should we put comment describing base_name in tasklist.proto? it will be hard to find this in the future

imo yes. Having a well documented IDL is a really good multiplier whenever someone is interacting with our interfaces (e.g building additional clients, building their own cli, etc.). It's a great place to put the expectations we have for each type, field, etc. in a single place. These idls are the definition of the expectations and behavior of our product so (if it is well documented) it becomes a great reference point. 

To your point it's less helpful for users of our thick clients (which is most people), but it can be a great starting point for the documentation at that layer too (or that can redirect them here).

proto/uber/cadence/api/v1/tasklist.proto
Shaddoll: I don't think it brings much value. Currently, the server just uses the name field to perform task matching. It's passed across different layers including history, matching and database. 

> we don't need to manage whether or not it is the sticky, ephemeral, or original name with helper functions

With your change, we still need to tell if it's sticky, ephemeral or normal and use the computed name to perform task matching.

PR #263, opened 2026-05-06
feat(schedule): Add Schedule structs and methods to Thrift IDL

YaweiZhang-930:
### Summary
Adds Schedule API types and service methods to the Thrift IDL (shared.thrift and cadence.thrift). This enables the Cadence go-client SDK (https://github.com/cadence-workflow/cadence-go-client/pull/1501) to expose Schedule operations through the same unified **workflowserviceclient.Interface** that all other Cadence APIs use, rather than requiring users to wire up a separate proto client.  

### Type of Change
- [ ] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)
- [x] Other (please provide detailed description)
We are adding Schedule structs and methods to Thrift to avoid a behavioral change in the Go client SDK, since the existing NewClient and NewDomainClient constructors both accept workflowserviceclient.Interface (Thrift-backed). Without Thrift IDL support, users would need to separately construct an apiv1.ScheduleAPIYARPCClient (a proto-specific type) just to access Schedule — breaking the established pattern. 

### Data Effect
- [ ] Does it change the data stored in database?   Schedule state is owned and persisted server-side. This IDL change defines only the wire protocol — no persistence schema is affected.                                                                                                                                       

### Detailed Description
This is copying & pasting the existing Schedule structs & methods from proto to thrift. Every field should match

### Impact Analysis
- **Backward Compatibility**: Fully backward compatible. Adding new methods to WorkflowService and new structs to shared.thrift does not affect existing clients or servers. Existing clients that do not call the new Schedule methods are completely unaffected. Old generated clients simply do not have the new methods and continue to work as before.

- **Forward Compatibility**: Schedule is available over two transport layers independently. The gRPC path is already fully implemented server-side — gRPC clients can use Schedule as soon as the server is deployed. The TChannel path requires an additional server-side change: a type conversion layer (common/types/mapper/thrift/schedule.go) plus a gowrap regeneration to produce the handler stubs — no business logic duplication, as the frontend handler and scheduler domain logic are shared between both transports. TChannel clients calling Schedule against a server that has not yet deployed the Thrift handler will receive an error response until that change lands.

### Testing Plan
- **Unit Tests**: This is an IDL-only PR.
- **Persistence Tests**: Not applicable — this PR does not change any persisted data types.
- **Integration Tests**: Integration tests will be added in the go-client and server repos once the server-side Schedule handlers are implemented and this IDL is consumed.  
- **Compatibility Tests**: Not required for this change — it is purely additive. Existing methods and types are untouched.    

### Rollout Plan
- What is the rollout plan?  N/a
- Does the order of deployment matter? NO
- Is it safe to rollback? Does the order of rollback matter? Yes. Since this is a purely additive change with no modifications to existing types or methods, rolling back this IDL has no impact on existing clients or server behavior. 
- Is there a kill switch to mitigate the impact immediately? Not required at the IDL level. The Schedule feature is gated by the server-side implementation.

### DIFF
--- a/thrift/cadence.thrift
+++ b/thrift/cadence.thrift
@@ -768,7 +768,9 @@ service WorkflowService {
       1: shared.BadRequestError badRequestError,
       2: shared.EntityNotExistsError entityNotExistError,
       3: shared.ServiceBusyError serviceBusyError,
-      4: shared.AccessDeniedError accessDeniedError,
+      4: shared.QueryFailedError queryFailedError,
+      5: shared.LimitExceededError limitExceededError,
+      6: shared.AccessDeniedError accessDeniedError,
     )
 
   /**
@@ -780,7 +782,9 @@ service WorkflowService {
       2: shared.EntityNotExistsError entityNotExistError,
       3: shared.ServiceBusyError serviceBusyError,
       4: shared.DomainNotActiveError domainNotActiveError,
-      5: shared.AccessDeniedError accessDeniedError,
+      5: shared.LimitExceededError limitExceededError,
+      6: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
+      7: shared.AccessDeniedError accessDeniedError,
     )
 
   /**
@@ -792,7 +796,9 @@ service WorkflowService {
       2: shared.EntityNotExistsError entityNotExistError,
       3: shared.ServiceBusyError serviceBusyError,
       4: shared.DomainNotActiveError domainNotActiveError,
-      5: shared.AccessDeniedError accessDeniedError,
+      5: shared.LimitExceededError limitExceededError,
+      6: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
+      7: shared.AccessDeniedError accessDeniedError,
     )
 
   /**
@@ -804,7 +810,9 @@ service WorkflowService {
       2: shared.EntityNotExistsError entityNotExistError,
       3: shared.ServiceBusyError serviceBusyError,
       4: shared.DomainNotActiveError domainNotActiveError,
-      5: shared.AccessDeniedError accessDeniedError,
+      5: shared.LimitExceededError limitExceededError,
+      6: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
+      7: shared.AccessDeniedError accessDeniedError,
     )
 
   /**
@@ -816,7 +824,9 @@ service WorkflowService {
       2: shared.EntityNotExistsError entityNotExistError,
       3: shared.ServiceBusyError serviceBusyError,
       4: shared.DomainNotActiveError domainNotActiveError,
-      5: shared.AccessDeniedError accessDeniedError,
+      5: shared.LimitExceededError limitExceededError,
+      6: shared.WorkflowExecutionAlreadyCompletedError workflowExecutionAlreadyCompletedError,
+      7: shared.AccessDeniedError accessDeniedError,
     )
 
   /**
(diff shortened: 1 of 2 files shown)


thrift/cadence.thrift
```diff
@@ -744,4 +744,102 @@ service WorkflowService {
       4: shared.EntityNotExistsError entityNotExistError,
       5: shared.AccessDeniedError accessDeniedError,
     )
+
+  // ── Schedule API ────────────────────────────────────────────────────────────
+
+  /**
+  * CreateSchedule creates a new schedule that triggers workflow executions on a cron spec.
+  **/
+  shared.CreateScheduleResponse CreateSchedule(1: shared.CreateScheduleRequest request)
```
abhishekj720, line 753: CreateSchedule is missing WorkflowExecutionAlreadyStartedError from its throws list?

thrift/shared.thrift
```diff
@@ -2292,3 +2292,202 @@ struct Predicate {
...
+struct ScheduleSpec {
+  // Standard cron expression (e.g., "0 6 * * *").
+  // Prefix with CRON_TZ to set timezone (e.g., "CRON_TZ=America/Los_Angeles 0 6 * * *").
+  10: optional string cronExpression
+  // Earliest time the schedule may trigger. If not set, starts immediately.
+  20: optional i64 (js.type = "Long") startTimeNano
+  // Latest time the schedule may trigger. If not set, runs indefinitely.
+  30: optional i64 (js.type = "Long") endTimeNano
+  // Random jitter applied to each trigger time to spread load.
+  40: optional i32 jitterInSeconds
```
abhishekj720, line 2328: In schedule.proto, jitter is a google.protobuf.Duration and catch_up_window is a Duration. Thrift uses jitterInSeconds and catchUpWindowInSeconds, that’s a deliberate simplification but it’s not a byte-for-byte copy of proto; sub-second jitter / windows can’t be represented. Either document that Thrift exposes seconds-only for these fields or align with how other Cadence Thrift types represent durations

thrift/cadence.thrift
YaweiZhang-930: No, because the handler explicitly catches it and converts it to BadRequestError:      https://github.com/cadence-workflow/cadence/blob/0c99a934268630e37df457b63bf79ce0f72b11ac/service/frontend/api/handler_schedule.go#L177                                                                                                                                         
```
  if errors.As(err, &alreadyStarted) {                      
      return nil, &types.BadRequestError{                                                                                                                                                                            
          Message: fmt.Sprintf("schedule %q already exists...", ...),                                                                                                                                                
      }                                                                                                                                                                                                              
  }   
```                                                                                                                                                                                                               
  Clients will never see WorkflowExecutionAlreadyStartedError from CreateSchedule.

thrift/shared.thrift
YaweiZhang-930: Good catch! The mental model for these changes are following the pattern: how every duration field in Cdence thrift is represented: executionStartToCloseTimeoutSeconds, jitterstartseconds and etc. I will add a comment to explicitly document these

thrift/cadence.thrift
YaweiZhang-930: I did another look at the cadence server, and I've updated the throw lists for schedule APIs.

- DescribeSchedule — added QueryFailedError and LimitExceededError
-  UpdateSchedule — added LimitExceededError  and WorkflowExecutionAlreadyCompletedError 
-  DeleteSchedule — same as UpdateSchedule:                                        
-  PauseSchedule — same                                        
-  UnpauseSchedule — same                                              
-  BackfillSchedule — same

abhishekj720 (approved):

PR #277, opened 2026-08-30
feat: Support Jitter on RetryPolicies

joannalauu:
### Summary
Add proportional jitter coefficient field on RetryPolicy

### Type of Change
- [ ] Add new API(s)
- [ ] Add new data type(s)
- [ ] Remove data type(s)
- [ ] Remove API(s)

### Data Effect

### Detailed Description

Added a jitter coefficient field, with changes as follows:
1. `proto/uber/cadence/api/v1/common.proto` and `thrift/shared.thrift` — `jitter_coefficient` field added to `RetryPolicy`
3. `thrift/sqlblobs.thrift` — `retryJitterCoefficient` added to `ActivityInfo` and `WorkflowExecutionInfo` to persist coefficient on retry

This coefficient defaults to 0, which means no jitter.

Part of [cadence #8238](https://github.com/cadence-workflow/cadence/issues/8238)

### Impact Analysis
- **Backward Compatibility**: Existing fields are unchanged. For existing table data, the new field will read as 0, which is no jitter, so behaviour is unchanged. 
- **Forward Compatibility**: Jitter is skipped for servers that do not know about this field, preserving existing behaviour

### Testing Plan

### Rollout Plan
- What is the rollout plan?


proto/uber/cadence/api/v1/common.proto
```diff
@@ -137,6 +137,11 @@ message RetryPolicy {
   // Either maximum_attempts or expiration_interval must be set.
   // Retrying will stop when either maximum_attempts is reached or expiration_interval is exceeded, whichever comes first.
   google.protobuf.Duration expiration_interval = 6;
+
+  // Coefficient for proportional jitter, used to spread retries out. Must be between 0 and 1.
+  // Each retry interval is multiplied by a random factor in [1 - coefficient, 1 + coefficient],
+  // never exceeding maximum_interval. Defaults to 0, meaning no jitter.
+  double jitter_coefficient = 7;
```
c-warren, line 144: I think a coefficient is the best option that we have for providing jitter to our users, as opposed to a raw value that they define. In terms of our options here, I think we have a few: 

1. An enum - e.g `NONE`, `FULL`, `DEFAULT` (default could be between 80-100% of backoff, like our client pollers). This would provide easier semantics to use, but wouldn't have a lot of control for the user. I like this as it reduced the amount of things a user needs to understand in order to use the feature. 
2. Symmetric coefficient aka this design. I think this is good except we need to ensure that the combination of jitter + backoff doesn't cause the retry policy to exceed the total expiration interval. This will be hard if we don't cap the jitter at 100% of the backoff interval, because we do calculations on the backend based on that + the maximum attempts. 
3. Downward-only coefficient - e.g the value will be somewhere between (t * (1-c) - t). This would allow control for the user, a default of 0 will be no jitter (which preserves backwards compatibility), and it could allow some sensible values (e.g we could cap c at 0.2, which would mean backoff is always 80% of the total backoff w/ jitter. 
4. Enforce it in the server only - I think we've ruled this out but it would be nice if everyone had jitter and just dealt with it - relying on specific retry interval durations in a distributed system w/ activities taking different amounts of time is already very brittle, so adding jitter to the retry backoff by default is imo very reasonable. 

@natemort wdyt? Any ideas here?

thrift/sqlblobs.thrift
```diff
@@ -151,6 +151,7 @@ struct WorkflowExecutionInfo {
   134: optional shared.CronOverlapPolicy cronOverlapPolicy
   137: optional binary activeClusterSelectionPolicy
   138: optional string activeClusterSelectionPolicyEncoding
+  139: optional double retryJitterCoefficient
```
natemort, line 154: Nit: Move this to be with the other retry policy related fields in the 80-100 range.

proto/uber/cadence/api/v1/common.proto
natemort: I think this (approach 2) generally makes sense as long as the behavior with maximum_interval still results in jitter. In particular, we need to make sure that we don't generate a jitter value and them clamp it to be less than maximum_interval as that would cause a significant bias in the jitter values generated and risks a thundering herd problem (exactly what we're trying to solve).

We need to do one of the following:
- Allow the value with jitter to exceed maximum_interval. This is fine as long as we document it.
- Don't allow the value with jitter to exceed maximum_interval, but generate a value in [1 - coefficient, 1] with all values having an equal probability.

I'm fine with either option (and the second matches the existing doc here), but we just need to keep it in mind. I have a slight preference for the first option.

c-warren: I'm concerned about 1), I think making maximum_interval a soft cap could be very unexpected for the user even with documentation. I'd prefer 2) where we do 1-coefficient, and ideally document that preferred values are between 0 and 0.2 (which keeps 80%+ of the retry backoff, minimizing the risk of a thundering heard caused by jitter).

joannalauu: I agree that 2) would be easier for users to understand, although it shrinks retry backoff. It might be more intuitive to implement @c-warren's suggested approach 1, where we give users 3 enum options: `None` (default, no jitter), `SMALL` (coefficient = 0.2), `HALF` (coefficient = 0.5).

I don't think that we should give users a `FULL` option, which I interpret to be coefficient = 1.0, because that could make the retry backoff really small.

c-warren: I like the enum approach generally, but I think 2) gives users more flexibility and means we don't need to extend the IDL when we have new requirements. 

We can also enforce a specific range at the server side (e.g max of 0.5), rather than in the IDL contract here.

joannalauu: I've updated the contract to note that we generate a value in [1 - coefficient, 1] and prefer coefficients between 0 and 0.2

c-warren (approved):

